PR #139, opened 2015-02-13
Make Vocaluxe buildable and playable on GNU/Linux

derivator:
Build with MonoDevelop or make.
Summary of the biggest changes:

* Correct for case differences between System.Data.SQLite and
  Mono.Data.Sqlite
* Exclude Direct3D and DirectSound from linux builds
* Provide linux versions of some windows-specific build events
* Exclude VocaluxeTests from linux builds so xbuild will work
* Add ability to build the native PitchTracker on linux
* Fix some problems with acinerella module
* Fix some windows-specific dependency checks that fail on linux
* Set OpenGL color format to 24-bit on linux


derivator:
Still needs a  linux equivalent of the powershell script that sets the assembly version in AssemblyInfo.cs. 
My previous [workaround for Mono's slow XML performance](https://github.com/derivator/Vocaluxe/commit/a1755757686f4a24298f5d3309fc48ddcdbd0847) is not included, as it seems to have become obsolete, probably because Mono improved their performance (or maybe Vocaluxe got faster).

xDarkice:
First of all, thanks for the effort making Vocaluxe compile on linux!

Mono is currently pulling in the XML stack from Microsoft's referencesource. That's probably why the XML implementation is much faster now.

On ubuntu 14.10 (should apply to 14.04 as well) you will need ffmpeg which is not in the official repos but can be pulled from ppa or installed from source. The same goes for mono from the packages which will fail compiling, but luckily there's a official mono ppa which pulls in the newest version which compiles fine.

I wasn't able yet to start up yet because OpenGL throws an exception but that's probably because I am on osx running ubuntu in an VM. Going to dig into that.

derivator:
Thanks to *you* for helping me with gstreamer-sharp and mono, you've been really helpful. And thanks to everyone on the team for making an awesome game and writing portable code that almost worked out of the box.

Since I don't have a mac and hate apple products quite passionately I can't comment on your issue, but it got me thinking, an osx version of Vocaluxe might be worth the effort, too, and should be in the realm of possibility, right?

xDarkice:
In theory a osx 32 bit build could work. Unfortunately, we rely a lot on System.Windows.Forms for input and the game window, which is not available on 64 bit osx. So we'd need an abstraction layer over WinForms for this.

xDarkice:
Got it running now. The newest OpenTK fixes my issues, so we should update that. Let me see if I can fix the GITVERSION thing using some sed magic

derivator:
Great! That would be appreciated. If you find any other problems, let me know.

derivator:
After some more testing, it seems that the web server is a problem for mono. Anything from trying to upload a profile picture to trying to listen to a song causes crashes. I haven't had a lot of luck trying to find the root cause yet...

xDarkice:
Btw, did you manage to start/compile Vocaluxe from monodevelop?
It might be worth trying with coreCLR though I am not sure how far Microsoft's work on linux is.

derivator:
Yes, it works fine with MonoDevelop.

derivator:
I've investigated the web server issues a little more. I believe most of them are caused by two mono bugs: [this](https://bugzilla.xamarin.com/show_bug.cgi?id=5926) one and [this](https://bugzilla.xamarin.com/show_bug.cgi?id=20764) one. I've posted to the mono mailing list and they're taking a look. The second one could be worked around in Vocaluxe, but I think for the first one we're dependent on a fix from them.

One issue turned out to actually be a problem in Vocaluxe: when a user uploads a new avatar, `CAvatar.GetAvatar` is called from the web server thread, which ends up calling `CDraw.AddTexture`, but that wants to be called from the main thread. On windows (with OpenGL) the avatar ends up blank, on linux Vocaluxe crashes. Any suggestions on how to handle this?

lukeIam:
@derivator: The webserver spawn a new thread for each request he gets.
We had some tries to make the connection the to main application threadsafe,
but we haven't found a perfectly fitting solution yet.

Anyway, I just did some research and I've maybe found a way to modify my solution to use the UI/main thread to execute the tasks:
https://github.com/lukeIam/Vocaluxe/compare/lukeIam:develop...ThreadSaveCServer
(Note: Only `GetProfileData` is using the new taks-model yet)

derivator:
That seems like a good idea. I've tried it (extended to SendProfileData) and it fixes the crash. I've made some further suggestions on your branch.

derivator:
Anything missing before this can be merged? Hope I don't seem impatient, just want to know what I can do to get this thing rolling :)

stsundermann:
The patch itself is fine, I'm ok with merging it once I confirmed it didn't broke the Windows build.

PR #142, opened 2015-02-28
Fixed PS script + Removed unused usings from CGamePad

lukeIam:
They seems to be a problem for mono


lukeIam:
Added: Fixed PS script for version injection into main.js

PR #144, opened 2015-03-01
Linux: set AssemblyInformationalVersion and fix CVocaluxeServer.GetServerVersion()

derivator:
@lukeIam could you check if this is doing what's intended? I'm not 100% sure I translated this: `%2524('@(VersionNumber)'+' ('+$version+')')` 
construction to `sed` correctly. My version produces output that looks like this:
`var matchingServerVersion = "0.3.0.83 (0.3.0.83-439-gacd89bc)";`

The previous code in `CVocaluxeServer.GetServerVersion()` was supposed to have 
`if (informationalVersion == null)`, right? Even then, on mono, converting it to a string would just give the classname, so I changed it.

### DIFF
--- a/Vocaluxe/Lib/Input/CGamePad.cs
+++ b/Vocaluxe/Lib/Input/CGamePad.cs
@@ -16,8 +16,6 @@
 #endregion
 
 using System;
-using System.Diagnostics.PerformanceData;
-using System.Drawing;
 using System.Threading;
 using System.Windows.Forms;
 using OpenTK.Input;
--- a/Vocaluxe/Vocaluxe.csproj
+++ b/Vocaluxe/Vocaluxe.csproj
@@ -536,10 +536,10 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent Condition=" '$(OS)' == 'Windows_NT' ">xcopy /E /D /Y "$(ProjectDir)\Website" "$(SolutionDir)\PhoneGap"
-powershell -command "$version = git describe --long; (Get-Content -Encoding UTF8 '$(ProjectDir)..\Output\Website\js\main.js') | Foreach-Object {$_ -replace '(?&lt;=var *matchingServerVersion * = *\\\")\w*(?=\\\" *;)', %2524('@(VersionNumber)'+' ('+$version+')')} | Set-Content -Encoding UTF8 '$(ProjectDir)..\Output\Website\js\main.js'"
+  <PostBuildEvent Condition=" '$(OS)' == 'Windows_NT' ">xcopy /E /D /Y "$(ProjectDir)\Website" "$(SolutionDir)\PhoneGap"
+powershell -command "$version = git describe --long; (Get-Content -Encoding UTF8 '$(ProjectDir)..\Output\Website\js\main.js') |  Foreach-Object {$_ -replace '(?&lt;=var *matchingServerVersion * = *\")\w*(?=\" *;)', "$version"} | Set-Content -Encoding UTF8 '$(ProjectDir)..\Output\Website\js\main.js'"
 powershell -command "(Get-Content -Encoding UTF8 '$(ProjectDir)Properties\AssemblyInfo.cs') | Foreach-Object {$_ -replace '(?&lt;=AssemblyInformationalVersion\()(.*)(?=\))', '\"GITVERSION\"'} | Set-Content -Encoding UTF8 '$(ProjectDir)Properties\AssemblyInfo.cs'"</PostBuildEvent>
-    <PostBuildEvent Condition=" '$(OS)' != 'Windows_NT' ">cp -ruT "$(ProjectDir)Website" "$(SolutionDir)/PhoneGap"</PostBuildEvent>
+  <PostBuildEvent Condition=" '$(OS)' != 'Windows_NT' ">cp -ruT "$(ProjectDir)Website" "$(SolutionDir)/PhoneGap"</PostBuildEvent>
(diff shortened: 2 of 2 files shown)


lukeIam:
> var matchingServerVersion = "0.3.0.83 (0.3.0.83-439-gacd89bc)";

The plan was this:
`var matchingServerVersion = "0.3.0.83 (7aebfe5)";`

But $version and @(VersionNumber) seems not to be fix - seemy comment: https://github.com/derivator/Vocaluxe/commit/71bd49d91659bf8b83f5b03a8d208ced19f8cdf6

Additionally something broke my injection mechanism, so atm nothing is not inserted.

derivator:
>I mentioned this problem before some time ago, but now we have to fix it
I don't understand the problem... `git describe --long` and `git describe --long --always` produce identical output for me, regardless of changes in the working directory. (git version 2.3.0)

How about we just go for
`var matchingServerVersion = "0.3.0.83-439-gacd89bc";`
that is, just use the unaltered output of git describe.

lukeIam:
> I don't understand the problem... git describe --long and git describe --long --always produce identical output for me, regardless of changes in the working directory. (git version 2.3.0)

Here is my output (git version 1.9.5.msysgit.0)
```
git describe --long
fatal: No names found, cannot describe anything.
```

```
git describe --long --always
7aebfe5
```
```
git describe
fatal: No names found, cannot describe anything.
```

> How about we just go for var matchingServerVersion = "0.3.0.83-439-gacd89bc";

As long that I can generate the same tag in the function `CVocaluxeServer.GetServerVersion` (with something like `Application.ProductVersion`) it's ok for me.
(I just need to check if client and server were build with the same version)

But this would effect the versioning scheme of the main program - so we might ask @stsundermann as it was made by him.

derivator:
That's very strange. From `git help describe`:
>The command finds the most recent tag that is reachable from a commit. If the tag points to the commit, then only the tag is shown. Otherwise, it suffixes the tag name with the number of additional commits on top of the tagged object and the abbreviated object name of the most recent commit.
[...]
--long
Always output the long format (the tag, the number of commits and the abbreviated commit name) even when it matches a tag.

So `git describe --long` would just output `0.3.0.83-439-gacd89bc`. I don't understand why it does something else for you.

>>How about we just go for var matchingServerVersion = "0.3.0.83-439-gacd89bc";

>As long that I can generate the same tag in the function CVocaluxeServer.GetServerVersion (with something like `Application.ProductVersion`) it's ok for me. But this would effect the versioning scheme of the main program

For me this is already exactly what [`Application.ProductVersion`](https://msdn.microsoft.com/en-us/library/system.windows.forms.application.productversion%28v=vs.110%29.aspx) outputs.

>ProductVersion first looks to see if the assembly containing the main executable has the AssemblyInformationalVersion attribute on it. If this attribute exists, it is used for both ProductVersion and CommonAppDataPath.

@stsundermann please confirm what the intended behaviour is.

stsundermann:
The git tag versioning is just for the nightly builds to make it easier to determine which build a user is running. In releases this version will be omitted.
So if ```git describe --long --always``` works then I'd go with this approach

```var matchingServerVersion = "0.3.0.83-439-gacd89bc";``` is fine

**Comment by lukeIam** [comment]
> @stsundermann please confirm what the intended behaviour is.

Strange problem - easy solution: my repo hasn't had a tag defined.
After pulling it the tags it looks good:

With and without working changes:
`git describe` ==> `0.3.0.83-442-g3c6dcb8`
`git describe --always` ==> `0.3.0.83-442-g3c6dcb8`
`git describe --long --always` ==> `0.3.0.83-442-g3c6dcb8`

And `Application.ProductVersion` also returns `0.3.0.83-442-g3c6dcb8` :smiley: 

@derivator 
so we use `0.3.0.83-439-gacd89bc` as version tag now? (that string is also displayed in the Vocaluxe application)
You would have to do some small adjustment here: https://github.com/derivator/Vocaluxe/commit/71bd49d91659bf8b83f5b03a8d208ced19f8cdf6
(`return Application.ProductVersion;`)

And I have to fix that powershell syntax/escaping error and the new version layout...

**Comment by lukeIam** [comment]
Just an idea: I do not link the use of the powershell commands in MSBUILD much.
What about making the the replacement with a MSBUILD custom task and some c# code?
(http://stackoverflow.com/a/22571621)

This should (?) also work for mono...

**Comment by derivator** [comment]
>What about replacing it with a MSBUILD custom task and some c# code?
Edit: it's not supported by xbuild: https://bugzilla.xamarin.com/show_bug.cgi?id=12171

That's too bad. What we could do is simulate this feature by adding a "build helper" project that outputs an executable that does our custom build steps. Then we could just do something like
```
<PreBuildEvent Condition=" '$(OS)' == 'Windows_NT' ">$(ProjectDir)\..\BuildHelper\BuildHelper.exe --pre</PreBuildEvent>
<PreBuildEvent Condition=" '$(OS)' != 'Windows_NT' ">mono $(ProjectDir)/../BuildHelper/BuildHelper.exe --pre</PreBuildEvent>
```
Pros:
* nicer to write than shell scripts
* we could do potentially very complicated custom build steps in a cross-platform way

Cons:
* it seems a bit overkill, we don't really have any extremely complicated custom build steps
* more projects

stsundermann:
If we want to to this properly then we'd use autotools for building. But the downside is the windows support as it's hard to setup a msys environment that works properly.
That's why I'd stay due to the lack of a replace xbuild target with the PS script on windows and use some sed magic on linux.

lukeIam:
ACK - would be a bit overkill.
I'll fix the PS code...

derivator:
I'll update this soon.

derivator:
Github won't let me reopen this so I'm going to create a new pull request.

PR #163, opened 2015-12-06
Increased default cover size (fixes #128)

### LINKED ISSUE #128: Low-resolution covers
Hello!

I noticed that Vocaluxe tends to load low-res covers when they are PNGs. I have a set of songs with 300x300 PNG covers, interlaced, and the UI only displays the degraded version of the picture. It doesn't look great.

Could this be improved?

lukeIam:


Flamefire:
300 is a bad value. Often it must be rounded to a power of 2. So 300px cover size will result in a texture of size 512px. Better choose 256 as the default. It is not much less than 300 but will save 4 times the memory

lukeIam:
Hmm your are right, but many covers are 300x300 on the net...
Every cover would be downscaled (maybe with contrast impact)

But the effect should't be that be -> 256 would be ok for me

2015-12-06 12:08 GMT+01:00 Alexander Grund <notifications@github.com>:

PR #164, opened 2015-12-06
Added option to stretch

### LINKED ISSUE #33: Support more screen resolutions
So far, Vocaluxe seems to only support 16:9 resolutions, adding black borders on any other kind of screen.
I'm using a 4:3 screen, and everything is scaled down and hard to read (tested on Win7 64bit).

(I don't really plan to use my 4:3 screens to actually sing, but move to a widescreen TV anyway, so that's quite a minor issue)

lukeIam:
Add an option to strech the rendered area instead of displaying black bars (closes #33)


lukeIam:
fixed

PR #175, opened 2015-12-19
Buzz Party Mode

Shiroi:
Implementation of the Party Mode described here: http://forum.ultra-star.de/viewtopic.php?f=93&t=11200


Vocaluxe/Lib/Input/Buzz/CBuzz.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            _Buzz.SetLEDs(led1, led2, led3, led4);
+        }
+        
+        private void _MainLoop()
+        {
+            _Buzz = new CBuzzLib();
+            _Buzz.BuzzChanged += _BuzzChanged;
+
+            while (_Active)
```
lukeIam: Just a thought: could this loop be replaced by a disconnect event (triggered from BuzzLib.cs:74)?
Atm this thread is doing nothing besides checking if the device has disconnected.
All work is done by your worker thread in BuzzLib.cs - maybe it can also handle the disconnect check.

ok it might require an additional event for changing _Active...

Vocaluxe/Lib/Input/Buzz/BuzzLib.cs
```diff
@@ -0,0 +1,229 @@
...
+            _Buff[4] = (byte)(led3 ? 0xFF : 0x00);
+            _Buff[5] = (byte)(led4 ? 0xFF : 0x00);
+            _Buff[6] = (byte)0x00;
+            _Buff[7] = (byte)0x00;
+            CHIDApi.Write(_Handle, _Buff);
+        }
+
+        private void _StartReader()
+        {
+            _Reader = new Thread(_ReaderLoop) {Name = "BuzzLib"};
```
lukeIam: I do not know how important it is: We always added `Priority = ThreadPriority.BelowNormal` for the Controller-Threads...

flokuep:
Nice enhancment... Quite don't know if it's (easily) possible: It would be nice, if we could use our Webserver/App also for voting the song... Just a thougt.

lukeIam:
Thanks for your PR! I would be a great feature for Vocaluxe!
Unfortunately I do not have Buzzers...

Vocaluxe/Lib/Input/Buzz/CBuzz.cs
Shiroi: Good Question...
I will have a look at this. This is nearly 1:1 copied from the Wiimote Class.

Vocaluxe/Lib/Input/Buzz/BuzzLib.cs
Shiroi: This could be done. Currently the Wiimote also didn't use the Priority at this point. Should this also be changed?

lukeIam:
> Quite don't know if it's (easily) possible: It would be nice, if we could use our Webserver/App also for voting the song... Just a thougt.

An additional "page" in the app with four buttons should be easy (even with covers images).
I just have to know which function I should call on data receive and how I figure out if there is a vote running (and with which songs).

What is not possible at the moment:
show the selection page automatically when there is a voting.
We do not have a channel back to the app.
But I'm planing to use the keep alive request to get events from the server.
Some light version of my old [try](https://github.com/Vocaluxe/Vocaluxe/wiki/HowTo:-Use-'PlayerCommunication'-(bidirectional-server-communication))
I think that attempt was a bit overkill (it could target single players and they could subscribe to events)

flokuep:
Hm. Maybe targetin single players (or player-groups) isnt' such an overkill... I could think about some party-modes, where just a team gets a survey.

Shiroi:
I've never looked into the Webserver...
This could be very interesting. Looks like I've got another thing to do :smile: 

To make this intuitive some kind of backchannel should be implemented.
It could also be used to let more players vote, like all players are voting or all except the singers

Vocaluxe/Lib/Input/Buzz/CBuzz.cs
lukeIam: @Shiroi You are totally right - Our Wiimote class does the same (haven't seen that, sorry) 

But I thing we cloud save one thread - the question is do we want to do it.
I can try to change it for the wiimote class - shouldn't be much work.
In case it is working we could adept it here...

lukeIam:
> I've never looked into the Webserver...
> This could be very interesting. Looks like I've got another thing to do :smile:
> It could also be used to let more players vote, like all players are voting or all except the singers

Just add your methods here:
https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/Base/Server/ICWebservice.cs
https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/Base/Server/CWebservice.cs
and the will be automatically exposed by the server. 

If you have question or need help don't hesitate to ask me.

> To make this intuitive some kind of backchannel should be implemented.

> Hm. Maybe targetin single players (or player-groups) isnt' such an overkill... I could think about some party-modes, where just a team gets a survey.

But backchannel  is an other topic: #176

PartyModes/PartyModeBuzz/Settings.StyleCop
```diff
@@ -0,0 +1,809 @@
+<StyleCopSettings Version="105">
```
lukeIam: Maybe this file should be removed or do I be at fault ?

Vocaluxe/Lib/Input/Buzz/CBuzz.cs
lukeIam: Hmmm - I tried to remove the additional thread and got stackoverflow exceptions in the underlying lib - maybe we leave it as it is...

lukeIam: I fixed the exceptions but I'm not sure if I've broke something else (I have no wiimote to test):
https://github.com/lukeIam/Vocaluxe/commit/d6d23809ecaa7625a52a4959584d36f2b56a4afa

PartyModes/PartyModeBuzz/Settings.StyleCop
Shiroi: Every Party Mode has a file like this one. It's a direct copy of the Challenge Mode I used as base.

Should I remove it?

lukeIam: > Should I remove it?

I have no idea...
@flokuep @stsundermann

derivator:
I have buzzers and will test this at some point in the (hopefully not too distant) future.

PartyModes/PartyModeBuzz/Settings.StyleCop
flokuep: Hm. Don't know. Maybe @Flamefire ?

Flamefire: Can't remember at least. Probably something really old...
I'd remove it. If we were to use a style then it should be global and therefore in the main folder

flokuep: Okay :D #1f695a7b83c59536740e9dc6b2dc5e80787a4cbb

Flamefire: Ups... :D
Well I can't remember anymore why I added those to the subfolders... Probably because the Partymodes are like separate projects to the styler... Maybe move them to a single new folder for styles which one can copy if one wants to use a styler

Shiroi:
@lukeIam I've changed the buzzers to only use one Thread based on your Wiimote test. It's working fine, so I added it to this PR.

Shiroi:
The HIDApi currently used isn't working on my Windows 10 PC. No devices are found.

I've compiled it from scratch and the new compiled version of the HIDApi works with Windows 10 & 7. Is it possible to update the HIDApi used by Vocaluxe?

lukeIam:
> Is it possible to update the HIDApi used by Vocaluxe?

I would just update the associated nuget package.
But first we should check if this breaks some stuff we are using (gamepad, wiimote?, ...)
(@derivator  do we use this lib also with Linux?)

And as you told we would have to compile it as the releases are pretty old...
@Shiroi if you send me the two hidapi.dlls (x86 and x64), I can create a nuget package and a prebuild test version (easy to test).

lukeIam:
> compiled it from scratch and the new compiled version of the HIDApi works with Windows 10 & 7

Question: As this is a VC++ project - which platform toolkit should I use?
Or in other words which VC++ redistributable should it depend on?

Or should should I just use static linking which should also solve this problem (no dependency to a VC++ redistributable)?
In that case the dll file size increases form 17KB to 89KB


*I vote for the static linking option.*

flokuep:
:+1: static linking

derivator:
@lukeIam yes, it's used on linux, too. Arch Linux packages version 0.8.0rc1 at the moment: https://www.archlinux.org/packages/community/x86_64/hidapi/

lukeIam:
> I've compiled it from scratch and the new compiled version of the HIDApi works with Windows 10 & 7. Is it possible to update the HIDApi used by Vocaluxe?

Please test:
https://ci.appveyor.com/project/VocaluxeBuildBot/vocaluxe/build/38

Shiroi:
A short test with my wiimote shows that it's not working on my Windows 10 PC. I'll have a closer look at this in the evening.

lukeIam:
@Shiroi which source code have you used to build hidapi?
I used d17db57b9d4354752e0af42f5f33007a42ef2906 from here: https://github.com/signal11/hidapi

Have you used a special version?

Shiroi:
@lukeIam I somehow deleted the source I used. The version from https://github.com/signal11/hidapi is not working for me. After some research I found this fork: https://github.com/cboulay/hidapi

It contains some Fixes for Windows 10 and I tested it on Windows 7 & 10 and it's working fine on both systems.

lukeIam:
> After some research I found this fork: https://github.com/cboulay/hidapi
> 
> It contains some Fixes for Windows 10 and I tested it on Windows 7 & 10 and it's working fine on both systems.

I overflew his changes - I will give it a try an create a new test version

lukeIam:
I've compiled your suggested version updated the packages and the test branch.
Please test: https://ci.appveyor.com/project/VocaluxeBuildBot/vocaluxe/build/45

Shiroi:
@lukeIam Just tested it and your version ist working fine on Windows 7 & 10.

PS: Sorry for the late answer.

lukeIam:
@flokuep what about this one?
I don't have buzzers to test.
As far as I can remember it's a combination of new input device + new party mode -> so should not be critical.

Maybe the theme file needs to be adjusted as we now have a new theme.

Off Topic: If we have a lot of free time 😆  it would be nice if we have a GUI that change depending on the input mode. So special gui modes for gamepad/Wii/Buzzer (color highlighting oder gamepad button icons of elements that can be triggered with them) = no longer plain mapping controller to keyboard keys.

lukeIam:
Merged in the current develop:
https://github.com/lukeIam/Vocaluxe/tree/pr_175

flokuep:
I don't have buzzers, too. Shouldn't be critical, but it maybe would be nice if we could have this party mode not only for buzzer-users. What do you think? How complicated would be having this in our webserver?

Thanks for merging current develop, will have a look at this (maybe later, I really don't know how much time there will be in december. Too much Advent and christmas stuff...)

flokuep:
I've thought about it again. Having voting for songs is a really nice feature for party modes, so I would like to have it in a more general way (like PartyNameSelection or PartySongSelection) so all party modes could use this with little effort. We can add this then as option to ChallengeMode. I don't think we should add new partymodes if there are not _really_ new and special and better try to integrate smaller features into existing party modes (otherwise: duplicate code, theme-stuff and translations).

Regarding webserver: Maybe we could have this without authorization and show a simple vote url for voting?

I've just scratched these to parts for better understanding:
page_2
vocaluxe_menu

lukeIam:
> We can add this then as option to ChallengeMode.
Would be the best choice - all other (and future) modes would profit from that option.




> Regarding webserver: Maybe we could have this without authorization and show a simple vote url for voting?

Would be nice - but I wouldn't introduce a new URL. Everything should be in one place/"app".
(Maybe we just add a login as guest button to the app)

OT: Furthermore would it be cool if we have more remote screens in the app for each screen shown in vocaluxe:
You activate a "second screen" in the app and it shows information of the current screen and allows to change them (e.g. player selection on the mobile).
But this would need us to redesign both app and Vocaluxe's screens to provide a kind off Status-And-Modification-Api.

flokuep:
> Would be the best choice - all other (and future) modes would profit from that option.

That's the plan. Shouldn't be too hard to implement, can do this.

> Would be nice - but I wouldn't introduce a new URL. Everything should be in one place/"app".
(Maybe we just add a login as guest button to the app)

Okay, maybe it would be nice to reconsider authentification in webapp. Passwords are hard to use and (maybe?) not necessary if user isn't allowed to control something. We could change this to an easy 'Who you are?' page (like Netflix after log in). If selected user has a special role (admin or something else) we can ask for a password... This might be easier to use especially for this kind of paty-mode voting.

> You activate a "second screen" in the app and it shows information of the current screen and allows to change them (e.g. player selection on the mobile).

YES! With this we can also bring second screen song selection during singing ;) Could you open a new issue for that so we have a nice place for planning? Or we could consider using GitHub's new(?) team discussion feature for that?

lukeIam:
> >You activate a "second screen" in the app and it shows information of the current screen and allows to change them (e.g. player selection on the mobile).

> YES! With this we can also bring second screen song selection during singing ;)

But with this I would suggest to recreate the web app -> it got a bit messy in the background :grin:
And the screens are note prepared for reacting to changes made somewhere else.
-> So we should think about a good concept first, as we could have different situations:
- "synced": e.g player selection open in Voacluxe and the second screen
- "Async": playing in Vocaluxe + second screen -> song or player selection 
- Let the user change between: also allow the user to see infos about the current song - maybe allow him to rate the players without influence on the real rating [show additional badges with "cool"/"second screen"/...] 

> Could you open a new issue for that so we have a nice place for planning? Or we could consider using GitHub's new(?) team discussion feature for that?

I'll open an issue. We can also start a team discussion (yes it's pretty new) to discuss but we should share our progress in the issue so that other can join if they like.

> Okay, maybe it would be nice to reconsider authentification in webapp.
We could change this to an easy 'Who you are?' page (like Netflix after log in). If selected user has a special role (admin or something else) we can ask for a password... 

Yes - I also think that the current system is to complicated (even I was the one created it 😄 )
I would also say that a 'Who you are?' page would be the best option. 

I also want to to have less roles: a normal role and an admin (maybe the admin can disable some normal role functions if he don't like them)

flokuep:
Okay. To sum up: We need to integrate this pull-request/voting-feature into ChallengeMode and find someone for testing. If someone has started doing this (maybe with this branch: https://github.com/lukeIam/Vocaluxe/tree/pr_175) we could close this pr and open a new one...

daggeg:
I don’t think I’ve paid any attention to this before. But this voting thing sounds pretty cool. I guess this ended up in ”developing hell” though, to use a Hollywood expression... I would otherwise like to try it :)

marwin89:
I close this pull request after 9 years. Supporting buzzer devices is out of scope.
I will add the basic idea to our [idea and wishes issue](https://github.com/Vocaluxe/Vocaluxe/issues/680) as

_Add song voting via Smartphone App in party modes._

PR #178, opened 2015-12-20
Ac multi thread wrapper

lukeIam:
Should fix our freeze problem on the song screen

### DIFF
--- a/Vocaluxe/Lib/Video/Acinerella/CAcinerella.cs
+++ b/Vocaluxe/Lib/Video/Acinerella/CAcinerella.cs
@@ -16,6 +16,7 @@
 #endregion
 
 using System;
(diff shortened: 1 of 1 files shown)


BenKmann:
Maybe the Concurrentdictionary can help?
Most Operations are Atomic and you don't need a second lock. The Dictionary itself is Thread-Safe
https://msdn.microsoft.com/en-us/library/dd287191%28v=vs.110%29.aspx#Anchor_10
and more modern, but also has some disadvantages. That's why it depends on the Situation. A long comparrison can be found here http://www.codeproject.com/Articles/548406/Dictionary-plus-Locking-versus-ConcurrentDictionar .

Sorry if this is already not an option because of some other rules.

Issue #169, opened 2015-12-06
Crash after starting a second challenge mode


mandark:
hi there!

first of all many thanks for your work on Vocaluxe! in my opinion it is the best and greatest singing game for the PC out there! it has provided me, my family and friends with countless hours of joy. i can't thank you enough for that! :D although it is already great as it is, i hope Vocaluxe will continue to improve and be maintained.

i'm having a particular problem with the challenge mode, and i think i've found a way to reproduce it. the problem seems to occur, if a challenge mode game has been completed, and a new one is started.

here are the steps to reproduce the problem:
1. start Vocaluxe
2. start a challenge mode (2 players and 1 round are enough to reproduce the problem).
3. you can abort the song immediately after starting it, since the error occurs also then.
4. the challenge mode is finished.
5. now start a new challenge mode.
6. the game crashes immediately after clicking "next" at the challenge mode options screen. you can find a screenshot of the error here: http://imgur.com/Y3bg6UC

after restarting Vocaluxe, a new challenge mode game can be started and completed without a problem. but when trying to start another go, the game crashes again.

i'm using the most current version i could find: Vocaluxe "Shining Heaven" v0.3.0 (x64) Beta (0.3.0.83-536-g378b9e6)
downloaded from here: https://ci.appveyor.com/project/lukeIam/vocaluxe/build/job/8dtvqdo1jax21g71/artifacts

but the error has been occuring since some other versions, i have tried.

maybe you can fix this problem. please let me know, if you need any more info.

many thanks and best wishes
gerald

flokuep:
Thanks for your report! I think this commit 50c19ff85df3e459e2d36c3b8d50aff6cd1a549e should fix it, still need to look into all changes by this dev and maybe fix one or two things...

mandark:
@flokeup
thanks for the info! i'm looking forward to the merging of this fix.

lukeIam:
Sorry - used wrong issue number first...
(ignore https://github.com/lukeIam/Vocaluxe/commit/4b9bfac1d9edc193c4d2a453082e94cc67966527 it is already removed from the branch)

mandark:
FYI: i've merged the changes from  z3ro01/develop to the current vocaluxe/develop branch in my local repo and compiled it. everything seems to be working fine and the challenge mode bug seems to be gone.

lukeIam:
should be fixed, or?
[...]

PR #182, opened 2015-12-21
Fix: Crash after starting a second challenge mode #169

z3ro01:


flokuep:
Thanks for splitting up your commits to branches and specially for this fix. Looks fine to me. :+1:

lukeIam:
No crash anymore - good!

But I'm experiencing an other behavior:
- I do have two songs in my Library.
- Start a challenge mode game (2 Players 1 Round)
- Exit the game after it has finished (song could be aborted)
- Start a new challenge mode game (2 Players 1 Round)
- There is **only one song visible** to choose (those which was not played in the first game)
- Exit the game after it has finished (song could be aborted)
- Start a new challenge mode game (2 Players 1 Round)
- There are two visible song again

I do not play the challenge mode - Is this a wanted behavior?

flokuep:
In general: yes. If you play partymodes several times, you don't have same Songs again and again. But yeah, we maybe should only do this if there are more than 10 (?) Songs...

lukeIam:
> In general: yes.

Ok then it is ok - It definitely makes sense...
I just thought new game - everything back to start.

> But yeah, we maybe should only do this if there are more than 10 (?) Songs...

Would be an option - but I don't think that there are may players with so few songs out there :wink: 

For the PR: good for me :+1:

PR #187, opened 2016-01-15
Wiimote single thread experiment

lukeIam:
Removes one out of the threads from the Wiimote handler.
One thread is just waiting that the other changes something -> introduced an event

I can not test this, but I think @Shiroi has tested it.
@Shiroi: can you please confirm...


Shiroi:
Just tested it again and the commit is working (on Win7). Wiimote is found and can be used to control Vocaluxe.

flokuep:
Okay, I guess we could merge then? :+1:

Shiroi:
:+1: for merging

lukeIam:
Thanks!

Issue #188, opened 2016-01-15
#PREVIEWSTART ignored for non video song


lukeIam:
The preview starts always from the beginning.

From here: http://forum.ultra-star.de/viewtopic.php?p=68969#p68969 (german)

PR #191, opened 2016-01-15
Smaller fixes

### LINKED ISSUE #189: Wrong number of total songs for folders
If you enter a song folder the total number of songs (right bottom corner) stills shows the number of all songs.

From here: http://forum.ultra-star.de/viewtopic.php?p=68969#p68969 (german)

flokuep:
This will fix #189 and #188


lukeIam:
Number of total songs in category: :+1: 
Preview-Start was ignored: :no_entry_sign: 

There seems to be an other problem with the Preview-Start:
Sometime it works mostly not.

flokuep:
Ah. This is not good. It seems that our PlaybackLib GStreamer couldn't read song-length as fast as we need it there....
Currently no good idea to solve this problem... (we may could get length while starting vocaluxe, but this will lead to longer loading times i gues...)

Shiroi:
What about caching the song length at first start? The initial start would be longer and all following ones could just use the cached values. Like it's done with the Covers (correct me if I'm wrong on this).

lukeIam:
> It seems that our PlaybackLib GStreamer couldn't read song-length as fast as we need it there....
> Currently no good idea to solve this problem...

Is there a possibility to figure out if GStreamer has already loaded the Length info (or even an event)?
If there is no event:
We could wrap all code after `_PreviewPlayer.Load(song)` to a task.
Then check on each render tick if the length is ready and then execute the task.

Not nice but loading at start time is also not really good.

lukeIam:
> What about caching the song length at first start? The initial start would be longer and all following ones could just use the cached values. Like it's done with the Covers (correct me if I'm wrong on this).

Would be a solution.
But if a song files changes it will completely fail - but i don't thing that that happens often

flokuep:
No, I think there isn't an event... Haven't worked with tasks and events yet...

lukeIam:
Other question: we just need the length greater than the file is long, right?

What would happen if we remove that check completely and let the player handle the problem that the position is out of the file.

flokuep:
I don't know if we could skip to a position if length isn't available anyway.

lukeIam:
Here is a workaround using tasks:
https://github.com/lukeIam/Vocaluxe/tree/previewStartBug

PR #202, opened 2016-03-26
Language revision

Jiiniasu:
Decided to check on the Swedish translation and ended up doing a bit more than that.

Apart from polishing the Swedish language file I've done the following:

* Made the lyric styles translatable
* "Fixed" the other two "English" languages (I'd rather prefer removal of Asturian in its current state as it's got so very few strings translated...)
* Made some changes in the English language file in an attempt to add consistency and try to abide by writing rules. (Not a native English speaker, so the latter might not be completely accurate...)

What are your thoughts?

(Unfortunately I don't speak any of the other languages fluently enough to revise them as well.)


derivator:
Looks good. Commit messages could be a bit more descriptive, but that's just a nitpick. I like that you broke it up into self-contained commits. Haven't tested it, but I'd say :+1: for merging.

flokuep:
Yeah - Thanks for your work! I guess we have a small problem: Our translations are done via transifex.com - I don't know if we could upload a translation there as a file manually (for swedish) - English wouldn't be a problem, this is our main-language and transifex pulls our xml-files from GitHub...
So: Maybe you need to do the Swedish-Stuff again...

BenKmann:
Hey, i also work with Transifex and in my Project we can Upload Files for new Languages manually (and even automatically). So it should be no Problem for someone with Transifex Admin Rights.

Jiiniasu:
I don't mind pasting the Swedish translation into Transifex as well. Although I couldn't really make heads or tails out on how to contribute on Transifex when I tried earlier. I've at least managed to submit a request to join the swedish translation this time.

Do tell how/if we should carry on with this.

PR #204, opened 2016-04-10
PNG screenshots

Jiiniasu:
Been taking a lot of screenshots while making my Vocaluxe-theme and have found it cumbersome that Vocaluxe saves them as .bmp.

I quickly browsed the code and made these changes. It seems to work fine and I don't see any reason why Vocaluxe shouldn't save screenshots as .png-files.

What are your thoughts?


flokuep:
Thanks for your work! I guess there were some problems with transparency, so we switched to Bitmap. But this was some time ago (I think before we moved to GitHub :D). I will test it, but don't see any problems...

flokuep:
Okay, after 8 months i've tested it. Everything is fine!

PR #205, opened 2016-04-10
More customisable notes

Jiiniasu:
While making my theme I wanted to have really transparant notes and yet have them filled in properly when sung. It didn't go too well for me, so I decided to dig into the code and make it a bit more themeable.
- I've split the textures used for the note and for the fill-in of the note. This way you may have a plain boring unsung note and have it filled in however fancy note you want when sung.
- The unsung note's colour can be altered by setting the `<NoteBaseColor>` in the skin xml. Falls back to White if not set
- The note lines can be coloured the same way with `<NoteLinesColor>`. Falls back to Gray with 0.5 Alpha if not set.
- Note's middle texture now repeats instead of stretching and no longer relies on a width of 2x end-width.

What do you think of the changes?

I have currently applied the same colour as the static colour to the note line and note base colous in the different skins for the default theme. Personally I think it's a bit much of the colours (Red is very red everywhere, etc). Would you rather have them plain gray/white? Screenshots below for blue comparison.

**Before change:**
Blue Theme

**After change:**
Blue Theme with Change

Also, sorry for the somewhat messy commits in this PR.


flokuep:
Thanks for your 2nd pull-request today :)  Yeah, I like these changes. Also like the new blue note-bars. It would be nice if you could increase theme-system-version as there are changes to theme-elements.
:+1 for merging!

Jiiniasu:
I'm not entirely sure what would motivate the theme-system-version increase. The new `<SkinFillLeft>`, `<SkinFillMiddle>` and `<SkinFillRight>` ups the version of ScreenSing. Am I overlooking something?

flokuep:
Do we want to include this into next release? I would support this, I guess @Jiiniasu would like to have it for Genius-theme? We need to:
- [ ] merge last develop-changes
- [ ] Increase theme-version

We need to increase theme-version and not screen-version because there are changes that affect a theme-element, but not screen-elements. (We need to change ScreenSing, but just because theme-version was increased)

Jiiniasu:
Apart from updating the branch and upping the theme version I'd probably want to review the code (at least) once more and see if I have a better solution for the 1px gaps that were present. Not sure if I should update this pull request or if it should be closed for now and that I should open a new one when I've done my review.

It would be nice to have this for Genius-theme, but I don't see it as an immediate necessity.

lukeIam:
what about doing the release with the current feature level (the last release was way to long ago 😉 ) and merging this PR (and some of the other PRs) into `develop` just after the release (of cause after we have the go from @Jiiniasu )

I hope the release after that one will be a bit sooner 😉

Jiiniasu:
I've upped the theme version, applied a better fix for the 1px gap problem and adjusted the notelines. Unless I have forgotten something or there are bugs that needs to be fixed I believe this change is good to go.

flokuep:
@Jiiniasu Could you merge latest develop branch into your branch? After that we can merge this 👍

Jiiniasu:
That should do it I s'pose. =)

PR #210, opened 2016-09-04
Highscore Listing Styles + Adjustments

Jiiniasu:
Added Highscore Style option;
List all - List all highscore entries for the song, ie the same player can have multiple positions in the highscore
List best - List best scores from each player at each difficulty for each duet-track for the song

Edited the SQL query to sort the scores on score desc, date asc. Newer entries with the same amount of points as older entries will have a lower position.

Adjusted the second number in the highscore screen to be the same size as the other numbers...

What do you think?

(I am not 100% sure about the name of the setting, but for now I can't figure out a better name.)


Jiiniasu:
Hm...
Seems like my SQL query is behaving poorly (wrong order when using group by) after testing with a bit more data. Need to see if there is a nice solution to the problem before merging this.

flokuep:
Thanks for your work! I really like this.

I guess a problem could be, that you use grouping, but try to select columns, that are not in group. So: On selection it's unclear what values for Score, Date and id should be selected (if SQLite works like other sql-systems, but I guess so). You need some kind of aggregation to select maximum score. But this may not work for id and score - I currently don't know if there is a first-function for fields on aggregation. This should be enough because of ordering.

For your noted problem: You may wrap your sql-statement. So:
SELECT PlayerName, MAX(Score) AS _Score,  FIRST(Date), First(id), Difficulty, LineNr, FROM ( SELECT * FROM Scores GROUP BY PlayerName, Difficulty, LineNr AS tmp) ORDER BY _Score DESC;

But I'm currently not sure if this syntax is completly correct and will work with sqlite...

flokuep:
And a little suggestion: Please add a complete new method for selecting best score in CHighscoreDB and then check in CScreenHighscore which of them should be called. With this, getting scores is independend from current config and we could reuse your method eg. for song-selection-screen (Some user want to see best three player there).

lukeIam:
@Jiiniasu 
If you can describe me what exactly the result of the SELECT should be then I may can help.

lukeIam:
> SELECT PlayerName, MAX(Score) AS _Score, FIRST(Date), First(id), Difficulty, LineNr, FROM ( SELECT * FROM Scores GROUP BY PlayerName, Difficulty, LineNr AS tmp) ORDER BY _Score DESC;

wouldn't work when your inner query returns something like:
```
score | date | id
 400  |   6  | 15
 100  |   1  | 5
  0   |   10 | 1
```

result would be: `400 |  1 |  1`

Jiiniasu:
I had written the following query before going to sleep last night:
```
SELECT PlayerName, Score, Date, Difficulty, LineNr, id
FROM (
  SELECT *
  FROM Scores
  WHERE [SongID] = @SongID
  AND [Medley] = @Medley
  AND [Duet] = @Duet
  AND [ShortSong] = @ShortSong
  ORDER BY Score
)
GROUP BY PlayerName, Difficulty, LineNr
ORDER BY [Score] DESC, [Date] ASC
```

It seems to me it is working as intended. Only distinct combinations of player and difficulty are showing and the entries are the highest scores for each. Players with the same score will have their position decided on score date. (Not sure what happens when multiple players have got the same top score at the same date...)

Regarding @flokuep 's suggestion of a new method: Adding a parameter for style would do the trick, wouldn't it?

I'll test a bit more after dinner before committing the updated code.

flokuep:
Yeah, adding a parameter would work ;)

lukeIam:
@Jiiniasu 
Your query is still ambiguous: You use `Score, Date, id` (not in a aggregation function) in your outer SELECT although there are not in the GROUP BY clause.
So if there are multiple values for one of these columns in a single group (e.g. two ids) then it is although which one to use.
I don't know how sqlite handles this, but I know databases permitting such queries at all and others choosing a random/first processed item.

Short: it could work, but it depends on the implementation of the database engine.

Jiiniasu:
@lukeIam, many thanks for explaining the weakness of my query. The query I've implemented in the updated code became the following:

```SQL
SELECT
		sc.PlayerName,
		sc.Score,
		MIN(sc.Date) AS Date,
		sc.Difficulty,
		sc.LineNr,
		sc.id
	FROM
		Scores sc
	INNER JOIN
		(
			SELECT
				Playername,
				MAX(Score) AS _Score,
				Difficulty,
				LineNr,
				id
			FROM
				Scores
			WHERE
				[SongID] = @SongID AND
				[Medley] = @Medley AND
				[Duet] = @Duet AND
				[ShortSong] = @ShortSong
			GROUP BY
				PlayerName,
				Difficulty,
				LineNr
		)
	AS mc
		ON sc.PlayerName = mc.PlayerName AND
		sc.Difficulty = mc.Difficulty AND
		sc.LineNr = mc.LineNr AND
		sc.Score = mc._Score AND
		sc.id = mc.id
	GROUP BY
		sc.PlayerName,
		sc.Difficulty,
		sc.LineNr,
		sc.Score
	ORDER BY
		[Score] DESC,
		[Date] ASC
```

flokuep:
Ah. Looks good for me.

lukeIam:
@Jiiniasu sorry for beeing the bad guy :disappointed: 
But I think we still have a problem here: `sc.id` is not in the GROUP BY and not in a aggregation function, so it's ambiguous.

Let's assume this dataset:

| PlayerName | Score | Date | Difficulty | LineNr | id |
|------------|-------|------|------------|--------|----|
| foo        | 5     | 10   | 1          | 1      | 1  |
| foo        | 5     | 1    | 1          | 1      | 2  |
| bar        | 6     | 1    | 1          | 1      | 4  |
| bar        | 6     | 10   | 1          | 1      | 3  |

Result:

| PlayerName | Score | Date | Difficulty | LineNr | id |
|------------|-------|------|------------|--------|----|
| bar        | 6     | 1    | 1          | 1      | ?  |
| foo        | 5     | 1    | 1          | 1      | ?  |

And if we assume that sqlite is always taking the first entry (I don't know the exact behavior) we are getting wrong results:

| PlayerName | Score | Date | Difficulty | LineNr | id |
|------------|-------|------|------------|--------|----|
| bar        | 6     | 1    | 1          | 1      | 4  |
| foo        | 5     | 1    | 1          | 1      | **~~1~~**  |

Two options:

- Remove the `sc.id` if we don't need it
- Restructure the query and add an other sub-query (that's a tricky problem...)
- Maybe there is another option I do not see at the moment

lukeIam:
Maybe something like this (untested):

```sql
SELECT
    os.PlayerName,
    os.Score,
    os.Date,
    os.Difficulty,
    os.LineNr,
    os.id      
FROM
    Scores os      
INNER JOIN
    (
        SELECT
            sc.PlayerName,
            sc.Score,
            sc.Difficulty,
            sc.LineNr,
            MIN(sc.Date) AS Date      
        FROM
            Scores sc      
        INNER JOIN
            (
                SELECT
                    Playername,
                    MAX(Score) AS Score,
                    Difficulty,
                    LineNr                
                FROM
                    Scores
                WHERE
                    [SongID] = @SongID 
                    AND [Medley] = @Medley 
                    AND [Duet] = @Duet 
                    AND [ShortSong] = @ShortSong              
                GROUP BY
                    PlayerName,
                    Difficulty,
                    LineNr          
            ) AS mc          
                ON sc.PlayerName = mc.PlayerName 
                AND sc.Difficulty = mc.Difficulty 
                AND sc.LineNr = mc.LineNr 
                AND sc.Score = mc.Score      
        GROUP BY
            sc.PlayerName,
            sc.Difficulty,
            sc.LineNr,
            sc.Score
	) AS iq   
		ON os.PlayerName = iq.PlayerName 
		AND os.Difficulty = iq.Difficulty 
		AND os.LineNr = iq.LineNr 
		AND os.Score = iq.Score 
		AND os.Date = iq.Date  
GROUP BY
	os.PlayerName,
	os.Difficulty,
	os.LineNr,
	os.Score   
ORDER BY
	os.Score DESC,
	os.Date ASC
```

Jiiniasu:
@lukeIam With a slight modification to your query it seems to work. The query as-is pulled ids from all over the table when I tested. As I am far from a SQL-wizard I am pretty much clueless on whether my fix opens other issues or somehow is a discouraged method to solve the problem. I'm totally relying on half-blind trial/error and checking output. ;) This time I've also checked the query against a mysql database with the scores. Query below:

```SQL
SELECT
    os.PlayerName,
    os.Score,
    os.Date,
    os.Difficulty,
    os.LineNr,
    os.id      
FROM
    Scores os      
INNER JOIN
    (
        SELECT
            sc.PlayerName,
            sc.Score,
            sc.Difficulty,
            sc.LineNr,
            MIN(sc.Date) AS Date      
        FROM
            Scores sc      
        INNER JOIN
            (
                SELECT
                    Playername,
                    MAX(Score) AS Score,
                    Difficulty,
                    LineNr                
                FROM
                    Scores
                WHERE
                    [SongID] = @SongID 
                    AND [Medley] = @Medley 
                    AND [Duet] = @Duet 
                    AND [ShortSong] = @ShortSong              
                GROUP BY
                    PlayerName,
                    Difficulty,
                    LineNr          
            ) AS mc          
                ON sc.PlayerName = mc.PlayerName 
                AND sc.Difficulty = mc.Difficulty 
                AND sc.LineNr = mc.LineNr 
                AND sc.Score = mc.Score
        WHERE
            [SongID] = @SongID 
            AND [Medley] = @Medley 
            AND [Duet] = @Duet 
            AND [ShortSong] = @ShortSong 
        GROUP BY
            sc.PlayerName,
            sc.Difficulty,
            sc.LineNr,
            sc.Score
    ) AS iq   
        ON os.PlayerName = iq.PlayerName 
        AND os.Difficulty = iq.Difficulty 
        AND os.LineNr = iq.LineNr 
        AND os.Score = iq.Score 
        AND os.Date = iq.Date
WHERE
    [SongID] = @SongID 
    AND [Medley] = @Medley 
    AND [Duet] = @Duet 
    AND [ShortSong] = @ShortSong 
GROUP BY
    os.PlayerName,
    os.Difficulty,
    os.LineNr,
    os.Score
ORDER BY
    os.Score DESC,
    os.Date ASC
```

lukeIam:
@Jiiniasu You are right - we need those additional WHERE-clauses.
Looks good for me now :+1:

Jiiniasu:
@lukeIam Awesome. =)
I'll commit the changes tonight.

Jiiniasu:
I've committed the new code and also corrected a typo I found elsewhere.

PR #211, opened 2016-09-07
Mod: prefer shipped gstreamer over an installed version (closes #199)

### LINKED ISSUE #199: Change prioritization of GStreamer installations
Currently we are preferring a system installation over the version we ship with the (nightly) releases.
This can lead into problems if the system GStreamer installation is a unsupported version or is missing components (plugins).

If we change to preferring the version in the Vocaluxe lib folder we would have better control over the GStreamer dependency and would make Vocaluxe easier to use.

lukeIam:
should reduce confusion if some users have an old version or a version with missing plugins installed.
Those who really want to use a different version of gstreamer (although I don't see the reason) will know how delete the folder of the shipped version.

If I interpret @derivator 's https://github.com/Vocaluxe/Vocaluxe/issues/199#issuecomment-188369766 right, the Linux version is not affected.


derivator:
Yes, should be fine. Thanks for remembering to consider impact on linux btw :+1:

PR #224, opened 2016-10-18
"Select random" in party mode name selection broken

derivator:
Currently if the number of visible profiles matches the number of used profiles or if there are more slots than usable profiles, an invalid profile with id -1 is selected if you click "Select random". I think the code `if (_VisibleProfiles.Count == _UsedProfiles.Count)` was trying to check if all profiles are already in use, but the way I understand `_UpdateVisibleProfiles()`, `_VisibleProfiles` already contains only the actually visible (=selectable profiles). Am I misunderstanding something @flokuep ? This way seems to work correctly.


flokuep:
Yeah. Looks good - Thanks for your changes, these lines were some of my first bigger changes I made in vocaluxe, so there should be some room for improvements.
👍 for merging

Issue #231, opened 2016-11-22
Little scrolling bug in PartyModeChallenge*


lukeIam:
The `1)` stays at the same position while scrolling:
![image](https://cloud.githubusercontent.com/assets/5115160/20537698/5bd1837c-b0ee-11e6-90df-ed9617dde793.png)

lukeIam:
`_RoundsTable[i].Number.Text` looks good and is changed as expected...

lukeIam:
There is definitely an other object with `1)` but I don't know where...

flokuep:
`<Text Name="TextRoundNumber">` has `1)` as default value (lines 178ff in `PartyScreenChallengeMain.xml`. I guess we use this element to get style informations for this table, but it seems to be drawn? You could try to remove `1)` from theme file...

lukeIam:
> You could try to remove 1) from theme file...

You where right: that solves the problem. Thanks!

PR #234, opened 2016-11-25
GUI-Debugmode

lukeIam:
While searching for the reason for #231 I started to highlight elements of the GUI.
Later I didn't want to throw it away so build a small GUI debug mode out of it.
Maybe it could be helpful for someone.
(`#define DEBUG_GUI` to activate)

What do you think?

image


flokuep:
Nice! I guess this could be really helpful for debugging and creating themes. In my opion, this would be a good addition to our (basic) theme-editor? So you could enable it everytime without new compiling (this would be useful for theme creators?) - I guess it's the F1-key. You could move your HandleMouse-stuff into HandleMouseThemeEditor. For drawing-stuff there might be a THEME_EDITOR-flag or something else...

lukeIam:
@flokuep I've chosen the compiler-flag because I didn't want to slow down normal playing.
Also this is the first time I hear that we have a build in theme-editor 😄 
I will have a look on it.

marwin89:
I close this pr due inactivity after 8 years. Please re-open if needed.
Basically this is a helpfull addition and I would like to see it comming.

PR #235, opened 2016-11-27
Change gamepad's Abort/"Esc" button from "X" to "B"

lukeIam:
as "B" is common key for aborting something (in all other games I know).


lukeIam:
@flokuep Can you also do a short review here please?

lukeIam:
thanks for approving!

PR #239, opened 2016-12-12
Support the versioning scheme

lukeIam:
e.g. `0.4.0.beta1`

- Mod: New versioning scheme including type if type!=release
- Mod: Show the same version tag everywhere
- Mod: Automatically determinate release type depending on the new version-tag
- Mod: Automatically determinate CodeName
- Mod: Adjusted pre-/post-build scripts for Windows and Linux
- Add: Code-name for the 0.4. releases 
"Blue Sunrise" - we can change that if you have a better idea 😉
First I thought about names of singing birds but the last release was "Shining Heaven".

### DIFF
--- a/Vocaluxe/Properties/AssemblyInfo.cs
+++ b/Vocaluxe/Properties/AssemblyInfo.cs
@@ -27,7 +27,7 @@ using System.Runtime.InteropServices;
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("")]
 [assembly: AssemblyProduct("Vocaluxe")]
-[assembly: AssemblyCopyright("Copyright ©  2013")]
+[assembly: AssemblyCopyright("Copyright © 2016")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 
@@ -49,6 +49,6 @@ using System.Runtime.InteropServices;
 //      Revision
 //
 
-[assembly: AssemblyVersion("0.3.0.83")]
-[assembly: AssemblyFileVersion("0.3.0.83")]
+[assembly: AssemblyVersion("SHORTVERSION")]
+[assembly: AssemblyFileVersion("SHORTVERSION")]
 [assembly: AssemblyInformationalVersion("GITVERSION")]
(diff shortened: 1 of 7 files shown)


Vocaluxe/Properties/AssemblyInfo.cs
```diff
@@ -49,6 +49,6 @@
 //      Revision
 //
 
-[assembly: AssemblyVersion("0.3.0.83")]
```
lukeIam: some little thing that steps out of the line:
`AssemblyVersion` have to be numeric (and `AssemblyFileVersion` shows a warning if it's not) - so we cant use `0.4.0.beta1` here.

so I replaced:
`alpha`->`0`
`beta`->`1`
`rc`->`2`
(According to the ERevision enumeration)

But besides `AssemblyFileVersion` in the Exe's file details those numbers are not shown.

flokuep: Maybe we could write our FullVersionText into AssemblyConfiguration or AssemblyDescription to clarify these numbers?

lukeIam: Sadly, neither AssemblyConfiguration nor AssemblyDescription is shown in the Windows Explorer.  
We could use `AssemblyTitle` (and `AssemblyProduct`=`Vocaluxe`)

http://stackoverflow.com/a/23145471

But in any case we then have to get CodeNames from a (xml/json) file...

lukeIam: I will think about a solution till the release

flokuep:
Currently have so problems to compile with Visual Studio...

Error	CS7034	The specified version string does not conform to the required format - major[.minor[.build[.revision]]]	Vocaluxe	Git\Vocaluxe\Vocaluxe\Properties\AssemblyInfo.cs	52	Active

@lukeIam Do you have some ideas?

lukeIam:
@flokuep Yes, sounds like my code generated an non-numeric AssemblyVersion (which should not happen).
On which branch does it occur?

flokuep:
flokuep/Vocaluxe:develop (Just tried to code a bit...)

lukeIam:
@flokuep It works for me without any error 😕 

Currently I'm moving/rewriting the whole versioning stuff to the build scripts (to have everything in one place)...
But for now you can try to revert any change to `\Vocaluxe\Properties\AssemblyInfo.cs`

Issue #206, opened 2016-06-13
Crash when song browsing


skyli91:
I had a crash 4 times in a long night. It's not bad but i think this should be fixed. Sadly I didn't make any screenshots and only the newest log is saved i think, so the error log is gone.
Is anyone else experiencing these crashes? They are not so frequent.

"Shining Heaven" v0.3.0 (x64) Beta (0.3.0.83-638-gd01ac27)

![image](https://cloud.githubusercontent.com/assets/7442913/16004852/78ac55ee-3164-11e6-9121-0ef2290694b5.png)

"Unhandled exception at 0x00007FFABABDE6FC (ntdll.dll) in Vocaluxe.exe: 0xC0000374: Egy halom sérült (parameters: 0x00007FFABAC322B0)." 

from error.log:
"350) Error opening video file (can't find decoder): C:\Dalok\Japán\Egoist - Departures -Anata ni Okuru Ai no Uta- (TV)\Egoist - Departures -Anata ni Okuru Ai no Uta- (TV).mp4" 

And after restart, I can play the song with the video too, without crash. (the error happens randomly, not only on this song)

skyli91:
Can it be becouse i have no gstreamer installed on this pc? I thought it's no longer required.

JPGsProfil:
same problem here, crash during song browsing. Only some times and no specific songs. One day it crashed after 20 minutes, the other day vocaluxe ran 3 hours without crash. The Screenshot isn't from the last build, but the problem also occur in the current. Hope it helps to fix the problem.

![fehler1](https://cloud.githubusercontent.com/assets/4520514/16419646/cc3ac5c4-3d4e-11e6-83a0-2ed43127552b.jpg)
![fehler2](https://cloud.githubusercontent.com/assets/4520514/16419654/d16ef1be-3d4e-11e6-8f69-76c73d3a9c01.jpg)

skyli91:
Sorry for the late reply:

The ffmpeg error log for this video where vocaluxe crashed one time: 
"Unrecognized option 'Anata'.
Error splitting the argument list: Option not found"

Mediainfo:
Format                                   : MPEG-4
Codec ID                                 : isom (isom/avc1)
File size                                : 18.1 MiB
Duration                                 : 1mn 33s
Overall bit rate                         : 1 631 Kbps

Video
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L4
Format settings, CABAC                   : Yes
[...]

PR #248, opened 2016-12-20
Use new acinerella and ffmpeg (3.2) versions

### LINKED ISSUE #241: Sporadic crashes when loading video
Hello.

This is probably the source of the crashes reported in #206.

In one of my branches I've (in the most ugly way) made a custom build for a LANparty where one of the modifications was instant previews when browsing (removing the need of pressing enter to start preview of song). This lead to more frequent crashing of Vocaluxe logging the error **Error opening video file (can't find decoder): <filename>**.

I've been digging a bit and have come to the following conclusion:

Below line somehow, sometimes, seemingly at random, throws a nullReferenceException. Occasionally, it can be recovered by running the same line again. Other times, it won't.

`decoder = (SACDecoder)Marshal.PtrToStructure(_Videodecoder, typeof(SACDecoder));`
(https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/Lib/Video/Acinerella/CDecoderThread.cs#L277)

**Edit:** The _Videodecoder variable also seemed to have a valid pointer when it crashes. All I can say is that it wasn't 0.

Have anyone got any ideas on how to figure out a solution for this bug or any plausable theories on why this happens?

This crash has been observed mostly during previews when browsing songs, but has also occurred as early as when loading the intro videos.

lukeIam:
Updated the acinerella code * to support ffmpeg 3.2+ and adjusted the integration into Vocaluxe (manged wrapper).

Should give us some more speed and it seems that it also fixes #241 (and #206)

\* Rewrote some of already existing functions, because of at that time nobody knew where the source of our acinerella modification was.


lukeIam:
@derivator Can you check the new versions on Linux please? (I've updated Vocaluxe/Lib/Video/Acinerella/acinerella.(h|c) )

lukeIam:
I finally failed to satisfy the runtime dependencies with Ubuntu 12.04 ...

lukeIam:
I really tried to test this under Linux and invested many days trying to get (normal) Vocaluxe running with Linux - without success.
I'm at the point I suggest to not longer hold back this PR and making the way free for the release (and for all those new features PRs waiting for post release time).

I don't think that this PR will break Linux support and there are not that many Vocaluxe-Linux users as the entry barrier is high.

lukeIam:
@flokuep @derivator I would like to merge this PR to prepare a release.

derivator:
@flokuep If I remember correctly, it should *run* fine on linux, if your distro ships a recent enough ffmpeg so you can build it at all...

flokuep:
@derivator Okay, sounds nice. I've tried it multiple times without success, but okay ;)

@lukeIam I guess you could merge it.

Issue #245, opened 2016-12-15
Include Genius-Theme into Releases


lukeIam:
https://github.com/Jiiniasu/Vocaluxe-Genius-Theme

Of cause only if @Jiiniasu agrees 😉 

![image](https://cloud.githubusercontent.com/assets/5115160/21238177/4a95b9f4-c302-11e6-8f2e-a5bf6fa60d4f.png)

Jiiniasu:
@lukeIam I approve of this. ;D

Jiiniasu:
On a little more serious note, I believe the party modes layout/design have to be checked if this is to be an official part of Vocaluxe. The party modes looks a bit weird with my theme graphics.

flokuep:
Yeah, i think this theme Looks really nice. But I think we should drop ambient then... Supporting two different themes is too much work I guess.

flokuep:
Done, see #256

PR #256, opened 2016-12-29
Genius-Theme as Default

Jiiniasu:
Hello,
In response to #245 I've made a branch where the Ambient theme has been replaced with the Genius theme I've made. I have also updated to the most recent version of Genius I had in store.

What do you think? Any improvements suggestions?


flokuep:
Wow! Really beautiful :) Thanks for this!

derivator:
It does look quite nice, but I've also grown quite fond of the way Ambient looks... supporting both would be more work I guess, but maybe it's worth it? I don't know how many active users are left, but removing Ambient entirely might be disruptive to those used to it. If we don't want multiple themes, what's the point of even having a theme system? (user-maintained themes seems unlikely since they can break at any time during development)

I would keep ambient at least for the coming release, since it's not really any effort at the moment. If during continued development it becomes too much work to support two themes, we could still drop it.

Jiiniasu:
One of the issues I see with having multiple themes is that the party modes sort of have their own theme, but still not. They have their own xml files for the screens, but they use the skin resources from the theme in use. This issue applies basically any theme and its design consistency throughout the game... One will have to decide which theme (if any) will dictate the look of the party modes.

When I've been developing Genius as a separate theme I had made the decision to simply ignore the party modes as they would always look weird and deviating. (I and the people I play with only use the regular singing mode anyway. =P)

I believe that the theming will need some enhancements before multiple themes can truly work side-by-side without looking odd.

I don't mind the idea of keeping Ambient, but it would trigger my OCD if the party modes makes one or both inconsistent in design style. =P

flokuep:
You could add a Genius theme for every partymode. Default is just a fallback (and uses current theme colors and resources, so developing there doesn't cost so much time)...

Jiiniasu:
@flokuep Huh, the more you know... I'll have to try it out. If it works like that it would be fine having both themes side by side.

Jiiniasu:
I've added Genius Theme files to the Party modes and reverted the Default fallback theme. If we want to include Ambient it is only a matter of reverting https://github.com/Vocaluxe/Vocaluxe/pull/256/commits/d7807844aa017be37c9780384cf93abefd9a822b as well.

I'm still not too fond of the theming of the party modes as they use parts of screensong and got elements not using the positions of the xml-files, but at least specific theme files for the theme is a step to the right direction.

flokuep:
Maybe I could help you :) There may be some jokers in ScreenSong, more I can't remember...

askusbloodfist:
Genius imo if a beautifull work. Should be Default or part of the main pack imo GREAT WORK !

flokuep:
After merging your note bar changes, there are conflicts now ;) Also, you could think about rating bars. I'm currently working on these, need them for 'Hold the line' game mode...

Jiiniasu:
I'll have a crack on the conflicts. Now I can also update the Genius notes to the latest ones. =P

Could you describe what you mean with rating bars?

flokuep:
https://psmedia.playstation.com/is/image/psmedia/SingStarPopEdition_sc0021?$MediaCarousel_Original$

The green thing  just beyond the points... And there will be a big red X that is shown if a player is _out_,

Jiiniasu:
Ah, I guess I could just alter one of the existing assets to make a rating bar if such asset is needed.

I think I resolved the conflicts. Needed a second try because I'm such a git noob. =P

flokuep:
Thanks, I will review this (after ci build is done).

Jiiniasu:
Noticed I missed a few things, so hang on a bit more...

Jiiniasu:
I believe it should be fine now. =)

flokuep:
Thanks. There is still a problem:

- PartyModes doesn't work. There seem to be some wrong versions (see error.log)

And I've some suggestions 😅 

- Contrast between selected and not selected values in select slides is maybe a little bit slight... It's sometimes hard to identify selected value. 
- It's a bit hard to read options of song menu in song screen. A bit less transparency could work here.
- Could you spend a little bit work on cover theme again? It's not really colorful. Maybe add option to display cover? And coloring it a bit?
- Song menu could be improved a bit generally imho. I really like the black bar on top of most screens. It could also work on song screen for showing select song I guess. (maybe a bit more like name selection?) Maybe you could increase size of song info part to the full height. We could show some more informations here if you like (year, highscore?)
- Playlist control need really a lot of space, maybe use the same size as on name selection, I guess you could use some coloring there?
- I really like name selection, sing screen and score screen  😍 

Thanks for your work! @lukeIam I'd like to merge this and after that do some improvements because I need to add some elements to screens and don't like to do this twice.

lukeIam:
@flokuep I'm not the theme expert - but as log as we are not breaking things you can merge this and improve the other points later (the nightly build is a development snapshot and must not be perfect)

flokuep:
We should put theme videos into a package again I guess, and we need to update installer. But this can be done later.

Jiiniasu:
Addressed the version errors in the log and did some colour adjustments.

Regarding the song menu and playlist size I've had ideas on redesigning the song menu and make space for highscore display etc. Although that would have to come in a separate PR or this one might get really, really old by the time I manage to realise such a redesign. =p

Kinda dry on ideas on the cover theme at the moment, but will try give it a ponder. If you've got any more specific ideas I'll gladly hear you out.

flokuep:
Thanks. I will do another test, and if everything is fine, I'll merge.

lukeIam:
just a note: At least for me the new background video is really slow lagging

Jiiniasu:
@lukeIam That's interesting. It runs fine on one of my weakest laptops (Intel Core i5 520M with the built in Intel HD Graphics). Although I do remember having issues with lag when I ran a debug build on a quite powerful machine in the beginning of the year... Worked fine with a release build though.

But the issue you have is isolated to the background video?

lukeIam:
@Jiiniasu 
I have also an Intel HD4000 integrated GPU.
The test was done with a release build.
Other music videos play fine (also the old theme video was perfectly smooth after ~1sec)

Working music video of a song:
working

Jiiniasu:
@lukeIam The reason is most likely differences in bitrate and resolution. You can try [this file](https://github.com/Vocaluxe/Vocaluxe/files/1290494/BG_Video.zip) of which I have tried to match resolution and bitrate of the Ambience videos.

I noticed now that the files I made for Genius are with quite high bitrates. Perhaps I should render new files with different settings than I had originally used...

lukeIam:
> I noticed now that the files I made for Genius are with quite high bitrates.

That could be a reason 😉 

> You can try this file of which I have tried to match resolution and bitrate of the Ambience videos.

Working 👍 
But we don't have to adjust our videos to fit my CPU in power saver mode 😆 
We just should agree on a good ratio between quality an performance.

By the way: ( @Jiiniasu @flokuep )
we should not put big binary files (like videos) to git. If there are any changes to those files they are stored twice and blow up the size our git repository (even if we delete the files later, they stay in the history)

flokuep:
Yes, I've missed this.. Could you add it as nuget-package?

Jiiniasu:
I'll prepare some new renders and run them through a proper encoder (x264). Please let me know how/where these should be stored to prevent the bloating of the repository.

lukeIam:
> Please let me know how/where these should be stored to prevent the bloating of the repository.

Just put the videos into a zip (with a folder Structure like `Themes\Genius\video.mp4`) and put it here:
https://github.com/lukeIam/VocaluxeDependencies/tree/master/zips/themes

I will create a small nuget packet which downloads the files from there (if they do not exist locally).

lukeIam:
BTW: Should I remove the Ambient theme from the build or do we keep it?

flokuep:
We could remove it.

Jiiniasu:
Encoded some videos in 1280x720 at 60 fps that can be tested.
Link: [Genius-Videos-720p60.zip](https://github.com/Vocaluxe/Vocaluxe/files/1293841/Genius-Videos-720p60.zip)

What do you think?

PR #263, opened 2017-01-08
Fixed markdown in posts

lukeIam:
Before:  
image
After:  
image

### DIFF
--- a/_posts/2012-07-04-first-final-release.md
+++ b/_posts/2012-07-04-first-final-release.md
@@ -7,7 +7,8 @@ author: "flokuep"
 image: release-01.png
 ---
 
-Dear Vocaluxe fans,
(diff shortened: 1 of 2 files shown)


derivator:
👍 of  course, but [didn't you already do this work](https://github.com/Vocaluxe/VocaluxeWebsite/commit/7b25ad67d8dafadeead026288388ce009d371386) in the VocaluxeWebsite repository? If we decided to go with the gh-pages branch approach, maybe we should merge the unmerged changes from there (if there are any) and delete that repository? Otherwise I will get confused by it many more times :)

lukeIam:
> of course, but didn't you already do this work in the VocaluxeWebsite repository?

Yes - I think they got lost while merging(?)

Issue #240, opened 2016-12-13
Homepage overhaul


flokuep:
Before realsing, I would like to give our homepage a complete overhaul.

Currently working on a new page with jekyll, current version: [flokuep.github.io/VocaluxeWebsite](http://flokuep.github.io/VocaluxeWebsite) - Once finished, we'll merge it into [Vocaluxe/VocaluxeWebsite](https://github.com/Vocaluxe/VocaluxeWebsite). I've finished the basic structure of our new page, there are still some tasks to do

- [x] Review all texts
- [x] Add some missing texts and screenshots
- [x] Missing content?
- [x] Find a way to set `JEKYLL_GITHUB_TOKEN` ([Read more](https://github.com/jekyll/github-metadata)) - We need this for automated download-content (currently there is 'no release')
- [x] Maybe some cleanup
- [x] Fix uncorrect unordered list in blogpost
- [x] Lightbox for images

derivator:
👍 for an overhaul. Just wanted to mention the possibility of going even simpler with [Github Pages](https://pages.github.com/). Maybe it would be nice to consolidate and ditch Sourceforge altogether? Github Pages also supports [Jekyll](https://help.github.com/articles/about-github-pages-and-jekyll/) for completely static sites, no php or anything involved. Just a thought.

flokuep:
Yeah. I guess completly switching is a good idea. Currenty don't know who is the owner of vocaluxe-domains to let them redirect to Github Pages (but this should be working sometimes).... @lukeIam has begon some testing with Jekyll and Github Pages. I guess it would be nice to move...

lukeIam:
> Currenty don't know who is the owner of vocaluxe-domains to let them redirect to Github Pages (but this should be working sometimes)....

| Domain       | Owner               | Active    |
|--------------|---------------------|-----------|
| vocaluxe.org | @brunzel            | Yes       |
| vocaluxe.net | @brunzel            | Yes       |
| vocaluxe.de  | Name + Post Address | No (typo) |
| vocaluxe.eu  | @stsundermann       | No        |
| vocaluxe.com | domain trader       | No        |

Even if it takes some time we can still serve a `301 permanently moved` from the sourceforge page and redirect all users.

P.S.: The facebook page uses vocaluxe.de -> so the link there is not working

@all
How should we proceed? Any design ideas?

flokuep:
[...]

PR #273, opened 2017-05-27
Small improvements on the webseite

lukeIam:
- Added a `read more` link to the news
- Optional manual truncation with `<!--more-->` and `<!--more-big-->`
- `usdb.animux.de` is currently offline
- example for the use of `lightbox` in markdown
- moved a image from sf to gh
- fix vertical align

fixes the most open points in #240


derivator:
👍

PR #278, opened 2017-06-15
Website Modern Design Overhaul

hendrikbl:
I updated the webpages design to a modern look: https://hendrikbl.github.io/Vocaluxe/
The overhaul also includes some mobile/responsive fixes and one or two phrasing changes to the sites text.


lukeIam:
@hendrikbl 
Thanks for your contribution! The site looks very nice!

I couldn't satisfy the new dependencies locally (problems with installing libv8 and therubyracer with windows), so I just had a look on https://hendrikbl.github.io/Vocaluxe/ .
Some small notes:

- Looks like the pre (beta) release is not detected
- We should think about an icon for medley (similar to the others)
- Are we allowed to use these icons? (license ?)
- hendrikbl -> vocaluxe in [_config.yml:L27](https://github.com/Vocaluxe/Vocaluxe/pull/278/files#diff-aeb42283af8ef8e9da40ededd3ae2ab2L27) and [_config.yml:L6](https://github.com/Vocaluxe/Vocaluxe/pull/278/files#diff-aeb42283af8ef8e9da40ededd3ae2ab2L6)

hendrikbl:
@lukeIam 
Thank you, I'm glad to hear you like it :)

I'm using the new "Bash on Ubuntu on Windows" feature of Windows 10 and I couldn't get it to run locally without therubyracer. Since github doesn't need it, I can remove it from the repo and use it only for myself.

Regarding your notes:
- The release isn't detected since it uses my fork of Vocaluxe which has nos releases except a nightly I created to test if it works. It sould work if the repo ist changed in the _config.yml.
- I agree, we need an icon but I just couldn't come up with a good idea...
- The party, duet and wiimote icons are made by me. The others are from https://materialdesignicons.com/ but without the fill color. So we can use them I guess.
- I forgot to change it again to the vocaluxe repo. I'll do that.

lukeIam:
> Since github doesn't need it, I can remove it from the repo and use it only for myself.

Without `therubyracer` it's working for me. Thanks!

> I agree, we need an icon but I just couldn't come up with a good idea...

Not easy at all...
hmm, maybe somthing like http://www.clipartdude.com/s.php?q=conference+call+icon with https://materialdesignicons.com/icon/music and inverse arrows??

hendrikbl:
> hmm, maybe somthing like http://www.clipartdude.com/s.php?q=conference+call+icon with https://materialdesignicons.com/icon/music and inverse arrows??

Great idea! I used it as an inspiration and came up with this:
Alt text

What do you think?

flokuep:
What about icon's licence? On https://materialdesignicons.com/ i've found this licence: https://github.com/Templarian/MaterialDesign/blob/master/license.txt; I guess we should put this into our image-folder, or not?

Otherwise: Nice works, thanks for your effort :) 👍

hendrikbl:
> I guess we should put this into our image-folder, or not?
It definitely wouldn't hurt having it in there 👍

PR #285, opened 2017-08-13
Fix: acinerella.dll is not statically linked 

### LINKED ISSUE #280: No video, audio only on some PC's
Ever since the builds with the new Acinella, I can only get Vocaluxe to work properly on 2 of the 4 PC's that I have tried it on.  On two of the PC's, the game runs but will not show video - either in preview or in the song. The Audio and everything else works fine.

I cannot figure out what the difference is or how to fix. 

I tried running https://github.com/lukeIam/VocaluxeProblemSolver but no luck. Does this need to be updated based on the changes that have been made the last couple months with video?

Old versions of the game work just fine on all 4 PC's.

lukeIam:
updates acinerella dependency (fixes #280)
Does somebody have a system without Visual C++ Runtime to test?


lukeIam:
@AlecScudder Can you confirm the fix?

AlecScudder:
@lukeIam How do I test?

lukeIam:
> I've just get the idea to try to run vocaluxe on a university remote desktop ;) Not sure if this is a good or a bad idea.

the overall target was to get rid of all other (VC) runtimes besides DirectX.

> VC++ 2012 Redistributables are missing. Please install them first.
> on start... So I guess no c++ runtime was installed there

Thanks! looks like there are additional dependencies to VC...
I will have a deeper lock on the dependencies of our dependencies.
(But acinerella.dll looks good for me...)

flokuep:
Okay. I'll be there for more tests.

Issue #277, opened 2017-06-09
Page is not updated on a new release


lukeIam:
The page vocaluxe.github.io/Vocaluxe/ is not recreated automatically on a metadata change (new nightly).
![image](https://user-images.githubusercontent.com/5115160/26868341-fd052ff4-4b68-11e7-8b67-fcea6bfbd268.png)

==> Do an empty commit after publishing in appveyor

Issue #276, opened 2017-06-06
BuildBot says latest nightly is from 2015


BenKmann:
Hi everyone,
you just merged some stuff and you're BuildBot created a new Nightly as expected.
But the Date is wrong: https://github.com/Vocaluxe/Vocaluxe/releases/tag/Nightly 
It says VocaluxeBuildBot released this on 7 Dec 2015.
Apparently he gets his Date from the Pull-Request start or something?
It looks a little strange on the releases Site ;)

lukeIam:
> It says VocaluxeBuildBot released this on 7 Dec 2015.

looks like this is the date of the first release ever of the buildBot...
atm I'm just deleting to old files and uploading new ones. Maybe we have to recreate the complete release each time - I'll put it on the list...

BenKmann:
What i found is, the only Page where you can see the real Date ist on the Tags Page https://github.com/Vocaluxe/Vocaluxe/tags there he knows that the nightly is only a day old.

lukeIam:
> What i found is, the only Page where you can see the real Date ist on the Tags Page

The tag is moved everytime [here](https://github.com/Vocaluxe/Vocaluxe/blob/develop/.appveyor/moveTagForSpecialBranches.ps1#L15)

But the [appveyor.yml](https://github.com/Vocaluxe/Vocaluxe/blob/develop/appveyor.yml) file deletes only the files of the nightly release and not the release itself. This was done to easily support appveyor (Windows build) and travis (Linux build) at the same time.

PR #286, opened 2017-08-14
Build script adjustments

lukeIam:
- Modified pre-release tagging: true for nightly and tags containing beta, false for all others (normal releases)
- Do an empty commit to gh-pages branch after publishing artifacts to trigger a github pages build -> fixes the outdated page issue #277
- Recreate the whole release on publish -> fixes the date issue #276 
As we do have to builds running parallel (x86 and x64) we have a race condition as the second could delete the release of the other.
I minimized the critical time by adding and checking current commit's sha1 tag to the release.
In my opinion the risk is not very high, but it can't solve the issue completely.
If we want to get sure [deployment environments](https://www.appveyor.com/docs/deployment/environment/) might help.

Edit: Sorry for mixing in f59393d (I can exclude it if required)


lukeIam:
> Maybe it would be nice (if possible?) to add APPVEYOR_REPO_COMMIT to gh-pages commit-message?

Sure 😃 See 03268ec

Finally I have say that I've tested the components of this separately on a dummy repo - but we have to double check if everything is working correctly together on the main repo. (please let me know when you like to merge it)

flokuep:
I've merged #285 - So feel free to merge this one and let us see i've everything works fine on our repo.

lukeIam:
> I've merged #285 - So feel free to merge this one and let us see i've everything works fine on our repo.

Releases seem to work 😄
But the dummy commit for gh-pages has some errors - I'm on it...

PR #287, opened 2017-08-14
Fix: Triggering a github pages update fails

lukeIam:
This should fix the errors during committing to the gh-pages branch


lukeIam:
@flokuep gh pages is now working (just some "errors" in the log but the are no errors)
But it looks like there's a problem when we:
- delete an old release
- create the new release
- it tries to delete a non existing asset (of cause there are no in the new release)
Mayby it's caching the asstes before the predeploy script is run??

See last lines of:
https://ci.appveyor.com/project/VocaluxeBuildBot/vocaluxe/build/117/job/bupxw92p94ah8ay0

For now I just re-run the build (second build does not recreate the release).
I have to investigate further on this...

Issue #281, opened 2017-06-28
Main screen - Preselect "Sing" after starting Vocaluxe, select "Exit" on ESC key press


Stefan1200-de:
Like the title says, in the main screen after starting Vocaluxe, please preselect **Sing**, so you just need to press on **Enter** to start the sing mode. And if the key **ESC** will be pressed while on the main screen, the selection should change to **Exit**. This would make the keyboard workflow much better.

Using Vocaluxe 0.4.0.beta1-31-g2e6b388 on Windows 10 x64.

PR #288, opened 2017-08-14
Smaller improvements

flokuep:
Resolved a bug (#268), added a small requested feature (#281) and pulled all language changes from transifex.

### DIFF
--- a/PartyModes/PartyModeTicTacToe/CPartyModeTicTacToe.cs
+++ b/PartyModes/PartyModeTicTacToe/CPartyModeTicTacToe.cs
@@ -60,7 +60,7 @@ namespace VocaluxeLib.PartyModes.TicTacToe
 
     // ReSharper disable ClassNeverInstantiated.Global
     public sealed class CPartyModeTicTacToe : CPartyMode
-        // ReSharper restore ClassNeverInstantiated.Global
+    // ReSharper restore ClassNeverInstantiated.Global
     {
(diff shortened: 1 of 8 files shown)


PartyModes/PartyModeTicTacToe/CPartyModeTicTacToe.cs
```diff
@@ -119,7 +121,7 @@ public struct SData
             public int CategoryIndex;
             public int PlaylistID;
 
-            public EGameMode GameMode;
+            public int GameMode;
```
lukeIam: Is there a special reason for changing EGameMode to int? (int would also allow invalid modes)

flokuep: Cast from selected value to EGameMode doesn't worked correctly beause of removing an option; Another way would be to add a custom enum for this game mode; We may change select slides sometimes to get selected enum values (I guess someone worked sometimes already on this 🤔 )

lukeIam: > Cast from selected value to EGameMode doesn't worked correctly beause of removing an option;

Understand - ok, I'll merge this.

> We may change select slides sometimes to get selected enum values (I guess someone worked sometimes already on this 🤔 )

This would be the best option for the future (I can remember something like this slightly)

PR #293, opened 2017-08-21
Filter out pre-releases that where released before the latest stable release

lukeIam:
@hendrikbl

### DIFF
--- a/_includes/download.html
+++ b/_includes/download.html
@@ -8,7 +8,7 @@
         {% assign release_nightly = release %}
     {% elsif release.prerelease == true and (release_pre == nil or release.created_at > release_pre.created_at) %}
         {% assign release_pre = release %}
-    {% elsif release_stable == nil or release.created_at > release_stable.created_at %}
+    {% elsif release.prerelease == false and (release_stable == nil or release.created_at > release_stable.created_at) %}
         {% assign release_stable = release %}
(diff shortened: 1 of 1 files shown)


hendrikbl:
Looks good but two things:

1. What qualifies a stable release as 'stable'? It looks like the code only sets `release_pre` and `release_nightly` but never `release_stable`. 

2. What about nightlies older than the stable release? Does this happen?

lukeIam:
> What qualifies a stable release as 'stable'? It looks like the code only sets release_pre and release_nightly but never release_stable.

You're right - the assignment was missing and the condition was incomplete.
(Hope this is now fixed)
We go through the following list and check the **bold** conditions (first rule matching is used):

1. Nightly: pre_release-flag + **tag name: "Nightly"**
2. Beta/Pre Release: **pre_release-flag** + tag name containing: "beta"
3. Stable: **no pre_release-flag**

> What about nightlies older than the stable release? Does this happen?

Should not happen: Nightly is always the latest commit on develop and the stable release is taken from develop. More precise: there can be a nightly build with an older publish date than the current stable release (actually it's very likely). But the will base on the same commit.

hendrikbl:
> (Hope this is now fixed)

It looks like it is so nothing more to complain from my side 👍

PR #303, opened 2017-09-02
Fix: mobile menu stays open after a navigation click

lukeIam:
-> close it after a click
#240


lukeIam:
@hendrikbl

hendrikbl:
Looks good 👍

PR #309, opened 2017-09-02
Rating bar

flokuep:
This adds a little rating bar as you can find in usdx and other similar games. We'll need this for (famous?) hold-the-line gamemode.

@Jiiniasu feel free to do create a better texture than mine (inverted one) 😄


Jiiniasu:
@flokuep Visually I'd probably want to make a white texture that can be coloured depending on how long the bar is (and put a static "frame" behind it).

Quick example:
ratingbar

Regarding the rating mechanics I get the impression that it is very forgiving. Perhaps it should be tweaked to be a bit more challenging?

As it is now I had to stay quiet for many lines (6?) before going low from having a full rating.

flokuep:
> Visually I'd probably want to make a white texture that can be coloured depending on how long the bar is (and put a static "frame" behind it).

Looks good. But quite not sure how this can be done with our current system.. We need to have Left and Right and Mid skin, and some different colors. Maybe I should introduce a new ThemeElement RateBar that could also be reused for ScoreScreen? What do you think?

> Regarding the rating mechanics I get the impression that it is very forgiving. Perhaps it should be tweaked to be a bit more challenging?

Yeah, feel free to propose some better calculation. Current one is quite dumb, but works for know. It calculations ratio of correct sung notes in current line and maps this to a value between -0.1 and +0.1. So if it is correclty sung -> +0.1, half correclty sung 0.0 and totaly missed -0.1 - All other is somewehere between. This is added to player's rating.

> As it is now I had to stay quiet for many lines (6?) before going low from having a full rating.

It should be around 10.

Jiiniasu:
I think a RateBar element would be a great, although I'd go for a more generic "ProgressBar" so it can be applied wherever a progressbar of some kind is needed in the future.

flokuep:
Will do that ;)

- SkinBackground
- SkinForeground
- SkinProgressLeft/Right/Mid
- Direction -> Up, Down, Left, Right
- AnimateMovement
- AnimateColoring
- ColorSteps (list of steps with different colors)

Anything else?

flokuep:
> Speaking of which, is it possible to make it have a single texture instead of separating left/mid/right? I've been pondering about textures sliceable to 9 parts for buttons and other rectangles to prevent "duplicate" textures to mitigate awkward resizing.

What do you mean? Do one texture and then specify in Skin.xml which part of this texture should be use?
`<SkinName>
    <Texture>...</Texture>
    <X>..</X>
    <Y>..</Y>
    <W>..</W>
     <H>..</H>
</SkinName>`

flokuep:
@Jiiniasu Could you prepare such texture so I can test with them properly? We may could reuse notes for this I guess?

Jiiniasu:
> What do you mean? Do one texture and then specify in Skin.xml which part of this texture should be use?

Sort of, specify one texture for the element and have it sliced up in the code.

Generic example of what I mean (for buttons/rectangular stuff):

**Texture**
btnexample1

**Same texture with slices outlined and numbered**
btnexample2

The texture should be sliced up in equally sized squares and follow the constraints below:
Slices 1, 3, 7 and 9 should never be stretched/repeated.
Slices 2 and 8 should only be stretched/repeated in horizontally.
Slices 4 and 6 should only be stretched/repeated in vertically.
Slice 5 should stretch/repeat in all directions.

If we had this kind of texture support we would be able to combine a lot of the current textures into one, making it easier to maintain the theme and quickly adjust sizes of rectangles without having to tweak textures.

In the case of the bars:

The source graphics file for the notes looks like this (with added background for visibility):
notes6

When making Vocaluxe textures out of it I slice it up in the 9 needed bits and save them separately. (In this case the end bits are half the width of the center bit.) We could have the texture fully combined like above, or have them separated in fill/base/bg.

flokuep:
Could you open a seperate feature request for this? Really like this idea, can't say how complicated this would be, but it would make handling of those elements a lot easier.

hendrikbl:
@flokuep @Jiiniasu We could use the same way of identifying the scalable parts as it's done in Android. See [NinePatch Drawables](https://developer.android.com/guide/topics/graphics/2d-graphics.html#nine-patch).

flokuep:
Okay, I've added a new progressbar element and use it for rating bar. Works quite nice so far. What needs to be done: Animations (and maybe some better theming? @Jiiniasu ).

Regarding NinePatch Drawables: Would be nice to have this, but I don't want to finish game mode stuff first... Also seems rather complicated to put this idea into our draw framework, so I need to think about it a lot before I could start working on this.

Jiiniasu:
I think the bar looks quite nice already. Although it doesn't seem to ever completely fill when I test it. There's always a bit left unfilled.

flokuep:
@Jiiniasu fixed it.

ProgressBar is animated now, Points on ScoreScreen are now displayed with ProgressBar element. So nothing left to do and I will focus on GameModes again ;)

Jiiniasu:
Little bugs I've found:
* Fading doesn't fade the fill of the rating bar
* 6 players with fading turned on gives the following error
```  Unhandled error: Index was outside the bounds of the array.   at Vocaluxe.Screens.CScreenSing._LoadNextSong()
   at Vocaluxe.Base.CGraphics._FinishScreenFading()
   at Vocaluxe.Base.CGraphics.Draw()
   at Vocaluxe.Lib.Draw.CDrawBase`1.MainLoop()
   at Vocaluxe.Lib.Draw.CDrawBaseWindows`1.MainLoop()
   at Vocaluxe.CMainProgram._Run(String[] args)
   at Vocaluxe.CMainProgram.Main(String[] args)
```

flokuep:
Thanks for testing; I've fixed both bugs.

Vocaluxe/Screens/CScreenSing.cs
```diff
@@ -846,6 +858,24 @@ private void _BuildStaticStrings(ref List<string> statics)
             }
         }
 
+        private void _BuildProgressBarStrings(ref List<string> progressbars)
```
lukeIam: Those `ref`-keywords can be removed or do they have a special function in this case?
We a passing a referencing a reference type and as long as we are not reassign (e.g. `progressbars = new List<string>()`) the parameter the `ref` should not have any impact.

Same for `_BuildStaticStrings()` and in `CScreenScore.cs`.

lukeIam:
OT: Any plan to add those Cool!/Poor/...-Popups too?
image

Vocaluxe/Screens/CScreenSing.cs
flokuep: Don't really think about them... Just copied. I've planned to spend some restructuring screen code (it's really nasty in some parts).

flokuep:
Yes, but maybe later... Having a proper animations system would be nice for that.

PR #311, opened 2017-09-06
Fixed drawing non-latin character not correctly, closes #310

### LINKED ISSUE #310: Problem with non-latin characters
When loading songs with non-latin characters, Vocaluxe can't display them properly. Charachters lay on top of each other.

I've already digged into this and found the problem: In CGlyph, L 54, MeasureString doesn't work properly and returns width=0, the same method works properly (without parameters) in line 50.

On StackOverflow I've found this:   https://stackoverflow.com/questions/11708621/how-to-measure-width-of-a-string-precisely This works (can't say why), but it breaks cover generation...

Any ideas what could be wrong and how to fix it (properly)?

flokuep:


lukeIam:
At least Chinese chars do not work:
image
image
Is this a font problem? Maybe we should have a fall back font.

flokuep:
Well, that's a bit strange. It's working on my system without problems. I've just tested it with different chars, copied them from wikipedia:




Could you run vocaluxe with FONT_DEBUG_OUTPUT (see CGlyph.cs)? This forces vocaluxe to save every single glyph as bitmap. File name should be the same as drawn character.

lukeIam:
> I've just tested it with different chars, copied them from wikipedia

Me too ... (pasted with notepad++ and UTF-8)

> Edit: Do you have the same problem on beta2, or just with my changes?

Also with beta2.

> Could you run vocaluxe with FONT_DEBUG_OUTPUT (see CGlyph.cs)? This forces vocaluxe to save every single glyph as bitmap. File name should be the same as drawn character.

There were some errors (missing variable and attribute) in the FONT_DEBUG_OUTPUT code - so I simplified it a bit.
-> the file names are also wrong...
-> The debugger show the same wrong text in CText.cs:
image

BTW: Are we rendering a texture for each char?

lukeIam:
Seems to be a parsing problem:
From Csongs::LoadSongs()

image
with:

```
#TITLE:無理碼  无理码
#ARTIST:Отречёмся от старого мира
```
And foldername `無理碼  无理码`

flokuep:
Okay. You might check CSongLoader.ReadHeader - Currently can't say where _Song.Encoding in instanciating stream reader is set - Could be a good start?

Could you also send me your .txt-file? I'm interested if i've same problems on my system with your file.

lukeIam:
> Okay. You might check CSongLoader.ReadHeader - Currently can't say where _Song.Encoding in instanciating stream reader is set - Could be a good start?

I just reached that line with the debugger 😆 
It's `iso-8859-1` and that will most probably cause the problems...
How do we detect the encoding?

The txt file:
https://gist.github.com/lukeIam/7c8f23657c971b6835009e08cd5e4d7e

flokuep:
Okay. So that's the problem... When #ENCODING is set in txt-file, everything should be fine (you might verify this).

Encoding is set in CSong, Line 91 (Encoding.Default).

lukeIam:
If I add this line to the txt it's working fine:
`#ENCODING: UTF8`

flokuep:
Okay. I've found the difference between yours and mine text-file: yours is utf-8 without BOM, mine with. If I've change this, everything works fine...

How should we handle this?

lukeIam:
> How should we handle this?

We could use something like this:
https://github.com/yinyue200/ude

But it will slowdown our song loading process...
Maybe as optional feature??

lukeIam:
OT: also the song loader reloads the whole file from disk if its finds an other ENCODING Tag. Thats also not very efficient.

flokuep:
It only reloads if it was red with a different encoding; But yes, still some room for improvements...

PR #313, opened 2017-09-12
Cleanup and party mode changes

flokuep:
* Removed outdatet theme loading stuff. This is safe now, because with (released) 0.4 old format could be converted to _new_ one.
* Added generic song config screen for party modes. This is most flexible now. You could choose between all songs, categories and playlists; also selction of song mode is possible.
* Remove ChallengeMedley because with new config screen we can also do this in Challenge.


flokuep:
Thanks for your feedback. I've fixed these "bugs" (and another one ;) )

Issue #320, opened 2017-09-26
Unhandled exception while quitting Vocaluxe


Stefan1200-de:
Now I got time to post this error here:
![image](https://user-images.githubusercontent.com/15168351/30880119-d72db67a-a301-11e7-8207-18f54b5a1ed6.png)

I had this error already with the Vocaluxe July 2017 nightly on quitting Vocaluxe. But not everytime, just maybe every third time I exit Vocaluxe. This time I had it with the beta 2 of Vocaluxe from early September 2017. Using Vocaluxe x64 on Windows 10 1703.

Error.log:
```
Error-Log Vocaluxe 'Blue Sunrise' 0.4.0.beta2 (x64) (0.4.0.beta2-0-g01fa98a) 26.09.2017 21:20:31
63) Unhandled exception: Es wurde versucht, im geschützten Speicher zu lesen oder zu schreiben. Dies ist häufig ein Hinweis darauf, dass anderer Speicher beschädigt ist.   bei PortAudioSharp.PortAudio.Pa_CloseStream(IntPtr stream)
   bei Vocaluxe.Lib.Sound.CPortAudioHandle.CloseStream(IntPtr stream)
   bei Vocaluxe.Lib.Sound.Record.PortAudio.CPortAudioRecord.Close()
   bei Vocaluxe.Base.CRecord.Close()
   bei Vocaluxe.CMainProgram._CloseProgram()
   bei Vocaluxe.CMainProgram.Main(String[] args)
64) Error during shutdown!: System.NullReferenceException: Der Objektverweis wurde nicht auf eine Objektinstanz festgelegt.
   bei Vocaluxe.Base.CBvideo.Pause(CVideoStream stream)
   bei VocaluxeLib.Menu.CMenu._PauseBG()
   bei VocaluxeLib.Menu.CMenu.OnClose()
   bei Vocaluxe.Base.CGraphics.Close()
   bei Vocaluxe.CMainProgram._CloseProgram()
```

System information:
[DxDiag.txt](https://github.com/Vocaluxe/Vocaluxe/files/1334718/DxDiag.txt)

Stefan1200-de:
Still occur with:
Vocaluxe 'Blue Sunrise' 0.4.0 (x64) (0.4.0-0-g41cf7a5)

Stefan1200-de:
I guess this is the crashdump about this issue. It is a 7zip file, but renamed to .zip, because I was unable to upload .7z files.

[...]

PR #325, opened 2017-10-04
PortAudio close-stream-handling

lukeIam:
Moved some checks to the code within the mutex.

Should fix the problem in #320 

Should be merged in `master` (to fix the bug in 4.0) and `develop`.


flokuep:
We should publish a bugfix-version 0.4.1, I guess?

PR #331, opened 2017-11-05
SEO for the webpage

lukeIam:
- Add Metadata
- Fix feed thumbnails
- Small config fixes

for #240 

@hendrikbl


_config.yml
```diff
@@ -12,7 +12,7 @@ url: "https://new.vocaluxe.org"
 #
 # NB! Without this *nothing* will work, because it's used in every path :)
```
flokuep: Not sure if this comment is for url or baseurl?

hendrikbl: I think as long as we don't use `{site.baseurl}` as a variable somewhere we should be good removing it. Referring to [this post](https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/) it's only used if the site is not hosted in the root directory of the server.

hendrikbl:
Looks good to me :+1:
I'm currently playing around with the styles and started to clean up the CSS and convert it to Sass. It should be way easier for future changes and readability.

_config.yml
lukeIam: looks like we have to set it to
```
baseurl: ""
```

Otherwise Github Pages sets it to `Vocaluxe`

PR #333, opened 2017-11-06
Option to autoplay previews

Jiiniasu:
This is a feature to let you enable autoplay of previews when browsing for songs. No need to press enter for a preview with this turned on!

Adds an On/Off option in Options > Game.

What do you think?

### DIFF
--- a/VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
+++ b/VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
@@ -64,9 +64,6 @@ namespace VocaluxeLib.Menu.SongMenu
 
         private readonly List<IMenuElement> _SubElements = new List<IMenuElement>();
 
-        private int _AutoplayDelay = 500;
-        private System.Timers.Timer _AutoplayTimer = new System.Timers.Timer();
-
         public override float SelectedTileZoomFactor
         {
             get { return 1.2f; }
@@ -578,38 +575,6 @@ namespace VocaluxeLib.Menu.SongMenu
             {
                 CBase.Songs.PrevCategory();
                 _EnterCategory(CBase.Songs.GetCurrentCategoryIndex());
-            }
-        }
-
-        private void _PreviewSelectedSong()
-        {
-            _PreviewNr = _SelectionNr;
-        }
-
-        private void _AutoplayPreviewIfEnabled()
-        {
-            if (CBase.Config.GetAutoplayPreviews() == EOffOn.TR_CONFIG_ON)
-            {
-                _PlayPreviewAfterDelay(_AutoplayDelay);
-            }
-        }
-
-        private void _PlayPreviewAfterDelay(double delay)
-        {
-            if (!_AutoplayTimer.Enabled)
-            {
-                _AutoplayTimer = new System.Timers.Timer(delay);
-                _AutoplayTimer.AutoReset = false;
-                _AutoplayTimer.Elapsed += (object sender, System.Timers.ElapsedEventArgs e) =>
-                {
-                    _PreviewSelectedSong();
-                };
-                _AutoplayTimer.Start();
-            }
-            else
-            {
-                _AutoplayTimer.Stop();
-                _AutoplayTimer.Start();
             }
         }
 
--- a/VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
+++ b/VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
@@ -364,10 +364,7 @@ namespace VocaluxeLib.Menu.SongMenu
                     if (_SelectionNr > 0 && moveAllowed)
                     {
                         _SelectionNr--;
-                        if (CBase.Config.GetAutoplayPreviews() == EOffOn.TR_CONFIG_ON)
-                        {
-                            _PreviewNr = _SelectionNr;
-                        }
+                        _AutoplayPreviewIfEnabled();
                         keyEvent.Handled = true;
                     }
                     break;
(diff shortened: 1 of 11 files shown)


Jiiniasu:
@lukeIam,

I've made an attempt in ading such a delay, currently set at 500ms. Feedback appreciated as I've just whipped this together after some self-education about timers through Google. =P

VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
```diff
@@ -586,6 +581,38 @@ private void _PrevCategory()
...
+            {
+                _PlayPreviewAfterDelay(_AutoplayDelay);
+            }
+        }
+
+        private void _PlayPreviewAfterDelay(double delay)
+        {
+            if (!_AutoplayTimer.Enabled)
+            {
+                _AutoplayTimer = new System.Timers.Timer(delay);
```
alraz1: No need for this. You have already created the object in line 68.
You should, however, set the Interval property of the _AutoplayTimer:
https://msdn.microsoft.com/en-us/library/system.timers.timer.interval(v=vs.110).aspx

I am assuming that there is a configuration file and that this _AutoPlayDelay value will be changed there.

If you have no plans of doing such thing, you should instead pass the "delay" parameter when you are creating the object on line 68 and remove this line

VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
```diff
@@ -586,6 +581,38 @@ private void _PrevCategory()
...
+                _AutoplayTimer.AutoReset = false;
```
alraz1: Same with this and all the properties below.
You should set them all once when creating the object (and not create the object every time you fire the timer =P)

The only thing that should be inside this "if" block is line 607, everything else should be on an initialization method.

VocaluxeLib/Menu/SongMenu/CSongMenuTileBoard.cs
```diff
@@ -60,6 +60,9 @@ class CSongMenuTileBoard : CSongMenuFramework
 
         private readonly List<IMenuElement> _SubElements = new List<IMenuElement>();
 
+        private int _AutoplayDelay = 500;
```
alraz1: You are basically copy-pasting the same code in both classes.
Instead of doing that, consider putting all the new functions you created in CSongMenuFramework (since both classes inherit from it) and only change the parts where you are calling your new functions. (for example, lines 201 )

VocaluxeLib/Menu/SongMenu/CSongMenuTileBoard.cs
```diff
@@ -549,6 +544,38 @@ private void _PrevCategory()
             }
         }
 
+        private void _PreviewSelectedSong()
+        {
+            _PreviewNr = _SelectionNr;
+        }
+
+        private void _AutoplayPreviewIfEnabled()
```
alraz1: this method is not called anywhere in this class

VocaluxeLib/Menu/SongMenu/CSongMenuTileBoard.cs
```diff
@@ -549,6 +544,38 @@ private void _PrevCategory()
...
+        {
+            if (CBase.Config.GetAutoplayPreviews() == EOffOn.TR_CONFIG_ON)
+            {
+                _PlayPreviewAfterDelay(_AutoplayDelay);
+            }
+        }
+
+        private void _PlayPreviewAfterDelay(double delay)
```
alraz1: neither is this (because it is only called by the above method =P)

VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
```diff
@@ -586,6 +581,38 @@ private void _PrevCategory()
...
         }
 
+        private void _PreviewSelectedSong()
+        {
+            _PreviewNr = _SelectionNr;
+        }
+
+        private void _AutoplayPreviewIfEnabled()
+        {
+            if (CBase.Config.GetAutoplayPreviews() == EOffOn.TR_CONFIG_ON)
```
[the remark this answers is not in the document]
lukeIam: Personally I also prefer always using curly braces (in my opinion it's safer when changes are done), but you are right most of the existing Vocaluxe code is following: "no braces unless needed". 

Maybe we should thing of a standard change - but this is an other topic...

VocaluxeLib/Menu/SongMenu/CSongMenuList.cs
```diff
@@ -64,6 +64,9 @@ class CSongMenuList : CSongMenuFramework
 
         private readonly List<IMenuElement> _SubElements = new List<IMenuElement>();
 
+        private int _AutoplayDelay = 500;
```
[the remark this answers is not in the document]
lukeIam: It can be added in CConfig.cs and without giving it a gui element (so it can be changed in the config.xml)

VocaluxeLib/Menu/SongMenu/CSongMenuTileBoard.cs
lukeIam: 👍 
Just move the functions to CSongMenuFramework  and change their visibility to protected.

lukeIam:
> I've made an attempt in ading such a delay, currently set at 500ms. Feedback appreciated as I've just whipped this together after some self-education about timers through Google. =P

@Jiiniasu Thanks for implementing me delay idea. 
As @alraz1 already mentioned you can reuse a single timer instance (faster than recreating new ones).
If I can help you with some of the changes please let me know.

@alraz1 Thanks for having a look on the changes and for writing good and detailed explanations!

Jiiniasu:
Thanks for awesome feedback! I've made changes based on the feedback.

flokuep:
Nice! Thanks for working on this @Jiiniasu - Just as a side note: Committing translations to GitHub (except english strings) will be overwritten when pulling new translations from transifex....

@lukeIam @alraz1 Thanks for reviewing!

Jiiniasu:
@flokuep,
Ah, good to know. I'll have to remember to update in Transifex then.

PR #337, opened 2017-11-19
Name Selection Improvements

Jiiniasu:
Fix: Infinite scroll bug with mousewheel
Change: Offset by row instead of page

What do you think?


lukeIam:
Got this when switching to the player selection screen:
image
image
(max = -3)

Jiiniasu:
@lukeIam Thanks for the feedback. Seems like I forgot to consider a profile count less than the amount of tiles.

Updated the branch and added a fix.

PR #338, opened 2017-11-23
Pause screen adjustments

flokuep:
I've improved pause screen a bit:

- Statics and text elements with Pause in its name are now automatically recognized and shown in pause screen (already merged in develop)
- There are new options to restart game or round
- Pause screen looks a bit cleaner now and buttons are better readable
- Code in ScreenSing was restructered for better readability


Vocaluxe/Screens/CScreenSing.cs
```diff
@@ -1440,184 +1389,301 @@ private float[] _CalcFadingAlpha()
...
+        {
+            for (int screen = 0; screen < CSettings.MaxNumScreens; screen++)
+            {
+                for (int numplayer = 0; numplayer < CSettings.MaxNumPlayer; numplayer++)
+                {
+                    for (int player = 0; player < CSettings.MaxNumPlayer; player++)
+                    {
+                        if (player <= numplayer)
+                        {
+                            string target = "P" + (player + 1) + "N" + (numplayer + 1);
```
lukeIam: Small thing:
`target` is never used (as far as I see) and cloud be deleted.
Also in line L1511 and L1574

flokuep: Ah. Yes. I've decided not to delete it because it's more readable with this var. But I've adapted it to use it correctly ;)

flokuep:
@lukeIam Some more problems? Would like to merge this as fast as possible, this could cause a lot of merge conflicts if someone touch ScreenSing...

lukeIam:
@flokuep go fo it! 😉

flokuep:
Can't ;) Need a review from you first :P

PR #339, opened 2017-11-23
GUID for profiles

### LINKED ISSUE #183: Unique Player-IDs
At the moment players has a incrementing player id that depends on the profile loading order (alphabetic?). Just to clarify: there is no id stored.

It would be good to have an id which stays the same over time and even over systems (maybe useful in the future).

Problem: there are so many places in the Vocaluxe code where the id is used for integer operations.
I tried four times to change the player id to a guid - but I screwed up each time.

So here is my suggestion:
give each player an **additional** guid next to the integer id (update existing profiles on load).

**Advantages**:   
- we do not break any existing code
- we create a doubtless possibility to identify a user if situation needs it
- we are creating a basis for a potential central profile system in the future

**Disadvantages**:   
- not that nice because we have two ids

flokuep:
This branch exists for a (very long) time - so maybe we could merge it now.

This will give profiles a guid instead of an internal integer id. With this, profiles are unique and can be shared/whatever. Especially this closes #183.

We can consider to save guid into HighscoreDB to have another reference to profile than name... Might be usefull for statistics-stuff or something else.


flokuep:
Nice, so we could merge it?

lukeIam:
Yes :)

PR #344, opened 2017-12-02
Fix: Button SelColor

Jiiniasu:
Correctly set Selcolor when creating one based on an existing button.


flokuep:
I would like to merge this into release-0.4.1-branch to have this fix available there, too... Could you apply this commit also on this branch?

PR #346, opened 2017-12-02
Fix: Button SelColor

Jiiniasu:
Correctly set Selcolor when creating one based on an existing button.


Jiiniasu:
For the release-0.4.1 branch as requested

PR #353, opened 2017-12-22
Distribute debug symbols with the releases 

lukeIam:
- Activation of pdb file generation
- Using [GitLink](https://github.com/GitTools/GitLink) to map the local source file path to the corresponding raw.github.com-file

You have to enable the `source server support` option in Visual Studio to be able to use the github files.

Can be used for debugging/profiling/crashdumps for release versions.

### DIFF
--- a/.appveyor/runGitLink.ps1
+++ b/.appveyor/runGitLink.ps1
@@ -1,4 +1,4 @@
-choco install gitlink --version 3.2.0-unstable0014 --pre --no-progress --limit-output
+choco install gitlink --version 3.1.0-unstable0017 --pre --no-progress --limit-output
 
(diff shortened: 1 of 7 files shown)


lukeIam:
Example build:
https://ci.appveyor.com/project/lukeIam/vocaluxe/build/job/vt4ekh677mau93kc/artifacts

flokuep:
What about adding this to 0.4.1, so we have this as fast as possible in a stable release version?

lukeIam:
> What about adding this to 0.4.1, so we have this as fast as possible in a stable release version?

Should also work there...
-> 👍 for cherry picking it to 0.4.1

Issue #364, opened 2018-01-05
Problems with "Control Vocaluxe with smartphone"


Michri:
When using last nightly build (37d5373 - 29.12.2017) I accessed Vocaluxe via iPhone (Port: 3000). Then I got problems:
1. First I tried to create profile via iPhone, but when clicking "Speichern und hochladen" I get an error: Bad Request . This happens with and without taking a photo upfront. I haven't logged in before.
BTW: Is a password required?

2. How can I login on smartphone with already defined users? What password?

3. "Nehme ein Foto auf und zeige es am Ende der Runde an" isn't showing any picture. What is wrong?

(Firewall disabled on PC.)

Michri:
Am I the only one having this issue?

lukeIam:
Hi @Michri,

the whole app/webserver needs a rewrite but the basic functions should work in most situations. 

> First I tried to create profile via iPhone, but when clicking "Speichern und hochladen" I get an error: Bad Request . This happens with and without taking a photo upfront. I haven't logged in before.
> BTW: Is a password required?

Can confirm this -> that whole profile creating/editing is broken atm.

> How can I login on smartphone with already defined users? What password?

Yes the users are the same and if no password is defined it's a blank password. As the passwords are stored salted and hashed there is no easy way to set a password without the profile edit feature.

> "Nehme ein Foto auf und zeige es am Ende der Runde an" isn't showing any picture. What is wrong?

Do you see the image in the Photos folder?

lukeIam:
The problems are caused by the change to GUID-PlayerIds...

PR #387, opened 2018-04-30
Use GUIDs in the app part

lukeIam:
Otherwise the app won't work correctly (see #364)


flokuep:
Not that big problem, but having these empty guid strings everywhere feels a bit ugly. I'd prefer defining a const value for this?

lukeIam:
> Not that big problem, but having these empty guid strings everywhere feels a bit ugly. I'd prefer defining a const value for this?

The whole structure of the App is a bit ugly and needs a complete rewrite 😉 
(Planed after redesigning the Vocaluxe-Classes -> after that it will be easier to access Vocaluxe's functionality).  
But nevertheless you're right -> I'll introduce a constant

lukeIam:
+ `IsProfileIDValid()` returns now false for `GUID.Empty`

PR #388, opened 2018-05-03
SonarCloud

lukeIam:
Run SonarQube analyses on every develop commit and on every pull request to find and track bad code style.


lukeIam:
SonarQube is a bit confised as there no analysis for the develop branch itself and it's complaining that we set "AssemblyVersion" to an invalid string ("SHORTVERSION").
-> We may should change this to something like "0.0.0" later.

lukeIam:
The "code-quality/sonarqube" check is configured as optional btw 😉

PR #392, opened 2018-07-01
Fix: Config.Video.VideoPreview is ignored (fixes #391)

### LINKED ISSUE #391: 0.4.0 and Nightly May 5 - Disabling video preview not possible
This issue was written in assumption that the setting **Video** / **Video preview** = **Off** should disable the video preview for the selected song in the song menu.

I tested Vocaluxe 0.4.0 and the newest Nightly from May 5, set all Video Options to Off. Disabling **Video backgrounds** and **Videos while singing** is working, no videos anymore. But after disabling **Video preview** I still get a video preview in the song menu after selecting a song. No matter if I use the Song-menu style **List** or **Tiles**, both have this issue. I also tested the Nightly from May 5 with default settings, I only changed the Video Options, same issue.

Does **Video preview** affect something different or is it really not working?

lukeIam:


Stefan1200-de:
Removed the change from @lukeIam and added the changes requested from @flokuep 
Tested them locally, seems to work fine.

PR #394, opened 2018-07-14
Fix: Proper Scaling

Jiiniasu:
Function scaled 2 * scale instead of scale.


flokuep:
Could you just explain shortly the problem with the current solution?

Jiiniasu:
@flokuep Sorry about the short description.

The problem with the current scaling code is that if you want to scale something it will incorrectly apply the scaling factor. For example setting `scale` to `1.5` would scale the rect by a factor of 2 since it will apply the width/height difference between the scales twice and offset the rect by the entire width/height difference instead of half of it.

See below table for a 1.5x scale test.

| Desc | W | H | X | Y | Scaling Factor |
|-:|-:|-:|-:|-:|:-|
| Input | 500 | 80 | 400 | 100 | 1 |
| Current Code Output | 1000 | 160 | 150 | 60 | 2.0 |
| New Code Output | 750 | 120 | 275 | 80 | 1.5 |

Stefan1200-de:
I started the current nightly and I started your Jiiniasu:ScalingFix branch. I was unable to see the difference. Do you have screenshots, where I can see the difference or can you describe what I should see?

Jiiniasu:
This PR only changes the behaviour of the scaling function as it was scaling items with an incorrect factor. Since I did not adjust any calls to it whatever is using the scaling will likely have its effect reduced. What immediately springs to mind if you want to look for differences would be the resizing of the album covers in the song select menu when hovered.

Screenshot_7
Screenshot_9

PR #400, opened 2018-10-22
improve song enumeration performance

swurzinger:
Currently a full directory enumeration is done before starting to load the first song. This commit changes the behaviour to enumerate the files lazily when loading the songs, so the song loading can start without a long delay at launch.


flokuep:
Thanks for your pull request :) Seems fine to me. Currently don't know what happend to our CI, I've just triggered it again.

Could you check if there are some more usages of `CHelper.ListFiles`, and if so replace it with `Directory.EnumerateFiles`? I think after this, we could remove `CHelper.ListFiles`...

swurzinger:
Instead of removing `CHelper.ListFiles` I modified it to use `Directory.EnumerateFiles` and return a `IEnumerable` instead of a `List` as its result is exclusively used for enumeration (in `foreach` loops) anyway.

I also fixed a small issue with the log message formatter that prevented placeholders from being replaced with their actual values.

PR #416, opened 2019-04-09
Relative song paths will now be converted to absolute song paths

GaryCXJk:
**Summary**

By setting relative song folder paths to absolute paths during initialization, you'll avoid issues with folder sort.

**Detail**

While using relative song folder paths isn't common, there are some use cases, for example, when setting Vocaluxe up on a portable storage device. In my case, I have both UltraStar Deluxe as well as Vocaluxe stored on an external drive, with the songs stored in the UltraStar Deluxe folder.

This usually isn't a problem, Vocaluxe can easily handle relative paths. However, there is a slight issue when you want to sort by folder, as Vocaluxe now tries to compare the relative path with the absolute path of each song. Since it is now unable to find a match with the relative folder, it will not be able to set the correct folder name.

So, as a quick fix, every relative path will now be converted to an absolute path. This will not modify the configuration file, so your setup remains portable.


Stefan1200-de:
> However, there is a slight issue when you want to sort by folder, as Vocaluxe now tries to compare the relative path with the absolute path of each song. Since it is now unable to find a match with the relative folder, it will not be able to set the correct folder name.

Have you an example for me, how I can reproduce that issue? Would be nice to understand and reproduce it, before I merge that change.

GaryCXJk:
I use Vocaluxe in a portable environment, where the drive isn't always static across computers. On top of that I load every song from a folder outside the Vocaluxe folder (Ultrastar Deluxe, in case I need to run that version). In this case, the song folder is set at `<SongFolder>..\UltrastarDeluxe\Songs</SongFolder>`.

After setting it up like this (though the issue could also be reproduced by using `.\Songs` as the relative folder), I put songs in different subfolders.

Stefan1200-de:
I will check that. In my "productive" Vocaluxe environment I have something similar (songs folder in a different game folder), but I only tested absolute paths. So I check that. If I think that your change has no other impacts, I will approve and merge it.

PR #435, opened 2020-01-10
Fix: x64 build (sonarcube problem with organization)

lukeIam:
Let's see if that helps...


lukeIam:
wrong base - sorry...

PR #436, opened 2020-01-10
Fix: x64 build (sonarcube problem with organization) 

lukeIam:
2nd try...


lukeIam:
Some Tests.VocaluxeLib.Log.CLogTest are failing but I think it's caused by the way the tests where written (needs some further investigations).

But this pr: LGFM

PR #439, opened 2020-01-10
Add small text line with key shortcuts to Song Screen, allow some key shortcuts while search bar is active

Stefan1200-de:
Add a small text line with key shortcuts to Song Screen, toggled if the search bar is active with help to the extended search keywords added by another commit. English and German text included in this commit.

Search bar invisible:
image

Search bar visible:
image

Party mode active:
image

**Small update related to this:**
On Song Screen the key shortcuts CTRL + R, CTRL + V and CTRL + S are now also allowed, if the Search Bar is visible. Especially CTRL + V is nice for this, because it allows you to sing all songs from the search result in random order. Example: You want to sing just songs with the genre rock? Do a search for g:rock and press CTRL + V. Have fun.

Small bug fix: CTRL + F is now disabled on song screen at party mode. It never really worked, but it was displayed for a short time.


Stefan1200-de:
Changed the separator in the German and English help text, so it looks a little bit nicer.

Stefan1200-de:
Added support for party mode song screens.

daggeg:
The additions of ctrl+A/ctrl+V are really nice. But to me they don't seem to be different as to their functions? Perhaps it's due to me a having the songs sorted in tiles instead as of a list?

PR #441, opened 2020-01-11
New github token for VocaluxeBuildBot

lukeIam:
Tokens for SonarCloud will follow when the have finished their maintenance.
(But SonarCloud tokens are configured in there portal and not as code/pr)


lukeIam:
Looks like something is not correct with the connection to the CI (no build are triggert)

lukeIam:
Going with OAuth again

PR #450, opened 2020-01-15
Fix: JumpTo not working correctly on folder sort (fix #449)

### LINKED ISSUE #449: Song screen - Selecting entry by pressing key not useful at folder sorting
If the song list is sorted for folder, currently on key press the _JumpTo() function jumps to the first entry of the folder path. Well, if the song library is on drive C:, only the letter c works and will select the first song on top of the list (not a big surprise).

Reason:
https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/Screens/CScreenSong.cs#L1098
I guess I should change
`element.Folder`
to
`element.FolderName`

But I want to check some stuff, before I do that and now I have to go sleeping. ;)

Stefan1200-de:


Stefan1200-de:
well, it's the same change, but for the nightly.

PR #458, opened 2020-01-18
Reload song from disk / key shortcut for skipping 30 seconds while singing

Stefan1200-de:
This changes will reload the current txt file from disk every time a song will be started (including lyric and notes). Also added key shortcut (F5 and CTRL + R) in pause menu on sing screen to restart the current song while reloading the song header (not the lyric or notes!). The restart game and restart round buttons from pause menu do also reloading the song header.

It is also possible to skip 30 seconds while singing using CTRL + Right Arrow.

This changes makes checking modifications to the song txt file much easier.

**Notice 1:** Reloading notes while singing is not working because some stuff is not working like I expect. But for checking new gap or start values this feature is excellent. But stop singing and start into the song again from song screen will reload everything, so it's still faster than restarting the whole game!

**Notice 2:** CTRL + Left Arrow is not possible, because bad things will happen. ;) But that is not so important, because you can restart the song from pause menu followed by CTRL + Right Arrow to get to the position where you missed something.

**Notice 3:** Song reloading from disk only happens for the normal game mode with no party modes enabled. From party menu no songs will be reloaded from disk. Same for medley or short game mode from song screen (since it will break medley or short game mode).

**Notice 4:** To disable the disk reloading stuff, just remove the song.reloadSong() line from CScreenSing.cs -> _LoadCurrentSong(). But please create an issue on GitHub, if you encounter problems because of this changes.


Stefan1200-de:
Currently reloading the song from disk have a small bad side effect if singing short versions of a song. I try to fix it today.

Stefan1200-de:
So, now I'm quite happy with this changes. Song reloading from disk now only happens for the normal game mode with no party modes enabled. From party menu no songs will be reloaded from disk. Same for medley or short game mode from song screen (since it will break medley or short game mode).

Stefan1200-de:
So, background image loading bug is also fixed. Looks good now.

Stefan1200-de:
Something is again broken with my local build. So this branch should not be merged, until I am sure everything is right.

Stefan1200-de:
Ah, my bad, VisualStudio just broke the new testing stuff with acinerella, has nothing to do with this branch.

PR #461, opened 2020-01-19
New Acinerella with ffmpeg 4.2 support

### LINKED ISSUE #381: Make Vocaluxe compatible with ffmpeg 3.3 / 3.4 / or newer
Vocaluxe so far only accepts ffmpeg 3.2. To have it working on multiple linux distributions it would be requiered to always make it compile against the newest versions of ffmpeg. In this case: ffmpeg 3.3 and 3.4.

So if someone is into this topic it would be nice to work on that.

Also, they made these adjustments for UltraStar Deluxe already. Maybe it's work having a look on the headers there. Thansk :)

### LINKED ISSUE #406: Crash with bad genre in video metadata
**_Behavior:_**
In the song selection, when the specific video start, it crashes Vocaluxe.exe with no message.
OS Logs tell crash occurs from acinerella.dll
The same song works well in UltraStar Deluxe.

**_Steps to reproduce:_**
1. Have an mp4 videoclip with metatag genre filled with:
`genre: -.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.--.-`
(add it with VLC for exemple)
2. Go in the song selection with video preview.
3. Select the "corrupted" song
--> freeze 1 second then back to desktop

**_Workaround:_**
As workaround we can remove the genre tag with VLC to make the videoclip working, but as it works well with Ultrastar, I think it could works in Vocaluxe.

lukeIam:
blocked by #462
fixes #381 fixes #406

Further steps:
- create a nuget package for Acinerella 
- create a nuget package for ffmpeg 4.2
- update [packages.config](https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/packages.config) file


lukeIam:
closes #462

lukeIam:
seems that we have some additional 3rd party packages (maybe to old) -> will try to fix those later today...

lukeIam:
@Stefan1200-de should be ready now - but I suggest that you have an other look on it (added the change to nuget.org + new packages for ffmpeg 4.2)

Stefan1200-de:
Downloaded that branch, VisualStudio was able to download everything. Confirmed that the genre crash is fixed, tested around 20 songs with video, switching aspect ratio, skipping 30 seconds multiple times, tested one song with background image and some other stuff. Everything looks good so far.

PR #691, opened 2024-04-18
Add ability to search by #TAGS

### LINKED ISSUE #690: Add ability to search by #TAGS
Hi,

I was excited to find out that Vocaluxe already has the support for the #TAGS attribute, but I was not able to figure out how to search using tags. Please correct me if I'm wrong, but searching through the source code, the capability does not seem to exist at present in the `develop` branch. Therefore, I would like to ask whether you are interested in adding the ability to search using the #TAGS attribute.

I'm new to the repo (and C#), but I'm willing to look into implementing it myself if you are OK with this feature (and you wouldn't prefer to do it yourself). Looking at the code, it should be fairly simple to implement as I believe all that is needed is to add a new case for the [`switch`](https://github.com/Vocaluxe/Vocaluxe/blob/cd0e5223cd320f12222ac703a6de0ce53bb4b5ed/Vocaluxe/Base/CSongFilter.cs#L118) statement in [CSongFilter.cs](https://github.com/Vocaluxe/Vocaluxe/blob/develop/Vocaluxe/Base/CSongFilter.cs) and then add a corresponding check similar to [that for genre](https://github.com/Vocaluxe/Vocaluxe/blob/cd0e5223cd320f12222ac703a6de0ce53bb4b5ed/Vocaluxe/Base/CSongFilter.cs#L207) in the for loop below.

RastislavTuranyi:
**Description**

Resolves #690 

The ability to search using the #TAGS attribute has been added, using `#` as the keyword in the advanced search. This was done by simply adding a new case to the search algorithm, mimicking that of the search by genre.

Currently, this new functionality does no appear in the help menu at the bottom: that would require updating the [language files](https://github.com/Vocaluxe/Vocaluxe/tree/cd0e5223cd320f12222ac703a6de0ce53bb4b5ed/Output/Languages) and I do not have the skills to update anything other than English. Furthermore, I'm unsure if an additional help would even fit on the screen. How would you like to handle this?

**To Test**

Please do have a look since I had to partially hack my local build of Vocaluxe to get it to compile and work at all.

1. Open the search using F3
2. Type in `#:` following a string of letters that match a #TAGS attribute in one of your songs; the matching songs should show up.

### DIFF
--- a/Output/Languages/en.xml
+++ b/Output/Languages/en.xml
@@ -155,7 +155,7 @@
   <string name="TR_SCREENSONG_NUMSONG">(%v song)</string>
   <string name="TR_SCREENSONG_ARTIST_TITLE">%a - %t</string>
   <string name="TR_SCREENSONG_HELPMESSAGE">A-Z = Jump to the letter • F3 = Search • CTRL+R = Random selection • CTRL+S = Start medley • CTRL+# = Change song sorting • CTRL+T = Enable/Disable category view • Space = Open song options • ESC = Back to main menu</string>
-  <string name="TR_SCREENSONG_HELPMESSAGESEARCH">A-Z = Search artist and title, prefixes for the search → a: = artist • t: = title • c: = creator • e: = edition • g: = genre • y: = year (e.g. 1999 or 1980-1989) • l: = language • al: = album • fi: = file name • fo: = folder name</string>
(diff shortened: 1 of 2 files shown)


flokuep:
Thank you very much for your contribution. You could test a ci build by yourself, just download it here: https://ci.appveyor.com/api/buildjobs/8yvxfx4937rims3l/artifacts/Vocaluxe_develop_Windows_x64.zip

Regarding the language file: adjusting English is fine. Vocaluxe resizes texts automatically, so I'd suggest just to try adding it.

RastislavTuranyi:
I have made the change for English, and everything seems to work fine using the CI build you shared from what I can see.

flokuep (approved):

flokuep:
Thanks for your contribution! If you like to do some work on vocaluxe, we would be very happy on it. Just have a look on our open issues, the most recent one are already tagged with some complexity notes and implementation hints. If you need any support, just ask :)

PR #704, opened 2024-08-18
Fix delayed perfect line effect

Jiiniasu:
A visual bug that has bothered me for a while where the sparkly effects aafter a perfect line gets delayed if the next line starts long after the perfect line ends.

Demo of current behaviour vs new behaviour attached. (We also seem to have the little helper dot off by an octave occasionally.)

https://github.com/user-attachments/assets/73ac79a9-a8f4-4340-9fc8-2d971cb5cc51


flokuep (approved):

flokuep:
Thank you for fixing that!

PR #719, opened 2024-11-05
Include VocaluxeProblemSolver by LukeIam

### LINKED ISSUE #200: Include VocaluxeProblemSolver into releases
Just an idea: How about including VocaluxeProblemSolver.exe in the nightly releases - so if someone is missing something it can be directly fixed.
http://forum.ultra-star.de/viewtopic.php?p=68551#p68551

marwin89:
## What does this PR?

- Include VocaluxeProblemSolver.exe by LukeIam
- This tool helps to solve some installation or microphone problems automatically.

## Why
Helps players to have properly installed and configured vocaluxe. It was never included since 9 years. It still works.

## Tested / QA
- ✅ tested successfully be myself (April 10th 2024)
-  It doesn't solve every problem but common errors

## More Info
- https://github.com/lukeIam/VocaluxeProblemSolver
- #200 

Description:
A little tool that deals with common Vocaluxe problems (for windows). At the moment it can detect and fix the following problems:
- No .NET Framework (4.0/4.5)
- No vcredist2010 x86
- No vcredist2010 x64
- No vcredist2012 x86
- No vcredist2012 x64
- No DirectX 9.0c
- No gstreamer
- gstreamer not fully installed (no mp3 codec)
- SingStar microphones configured to use only one channel (standard setting)
- Unblock blocked files


flokuep (changes requested):
I'm a little sceptical about merging the change like this. We add a binary (which is usually not a good idea in the Git system) and (theoretically) have further manual maintenance work. A better solution would be, for example, to provide the problem solver via NuGet or to download and pack it within the CI build in the same way as the theme videos. This way, we don't have any manual effort when the problem solver is further developed and the repository is free of external loads.

marwin89:
guess you're right 👍🏻

marwin89:
@flokuep : I would make a nuGET. 

Preconditions:

- [ ] 1. you transfer problem solver to an official vocaluxe repository in vocaluxe organization (lukeIam doesn't develop or maintain it)
- [ ] 2. you keep an eye on that repository as maintainer

Otherwise we skip it. I think problemsolver is helpful but I'm not keen on it 😄

flokuep:
Let's skip it. There is currently no one activley maintaining and developing Vocaluxe, so no need for kind of these changes. First real big maintenance task would be cleaning up and upgrading dependencies, and the problem-solver would become outdated...

PR #726, opened 2024-11-14
Additional fix for #725 (Fix Animation for Progress Bar on Score Screen)

GRaff844:
The Animation Timer also requires a reset. Otherwise the animation plays too quickly the second time the screen is displayed.

Apologies for the oversight in the initial fix.


flokuep:
No problem, thanks for fixing these things.

flokuep (approved):

Issue #675, opened 2024-01-20
Basic Sound Design


marwin89:
### What
Implement basic sound effects

- [x] On rating screen: Applause Sound: [Low](https://pixabay.com/sound-effects/small-applause-6695/) (2000-5000 Points) [Medium](https://pixabay.com/sound-effects/short-crowd-cheer-6713/) (5000-8000 Points) [Fantastic](https://pixabay.com/sound-effects/cheering-and-clapping-crowd-1-5995/) (8000-10000 Points)
- [ ] On rating screen: [Elevating Sound](https://pixabay.com/sound-effects/stuttering-climax-31002/) to illustrate pointsbar progress
(5seconds =100%, fade out sound earlier according to score)
- [ ] Warning Pop ups: [Info Chord](https://pixabay.com/sound-effects/point-in-space-36199/)
- [ ] Pop-Up Dialog: [Bloop](https://pixabay.com/sound-effects/bloop-1-184019/)
- [ ] Unlock Bonus: [Unlock Synth Sound](https://pixabay.com/sound-effects/fx-synth-80-itx27s-magic-177823/)
- [x] New Highscore: [Achive Sound](https://pixabay.com/sound-effects/achive-sound-132273/)

Sounds should be catchy but discreet and only on important game events. 

### Why
Improve immersion and fun for the players
Good singers are rewarded with sounds

### Additional Info
This task needs a programmer / sound programmer. As long as we don't have a dedicated sound designer ask marwin for sound effect research.

Edi316:
In my opinion, this issue can be closed:

Applause sound, Elevating sound, Warning sound, and Highscore sound have already been implemented.

There is currently no game mechanic for a Pop-Up sound or Unlock Bonus sound, so they are not applicable.

PR #727, opened 2024-11-16
Applause & Highscore Sound

GRaff844:
#675

- Applause sounds on rating screen, according to points achieved (Low Applause: 2000-5000 points, Medium Applause: 5000-8000 points, Fantastic Applause: 8000-10000 points)

If there is more than one player, the sound is based on the highest score among all players.

- Highscore Sound, if there is at least one new visible new entry on the highscore screen


Vocaluxe/Screens/CScreenScore.cs
```diff
@@ -250,6 +253,40 @@ private void _BuildProgressBarString(ref List<string> progressBars)
             }
         }
 
+        private int _ApplauseStream = -1;
+
+        private async void _PlayApplauseSound(int maxPoints)
+        {
+            // Play the appropriate applause sound based on maxPoints
+            if (maxPoints < 2000)
```
flokuep: We should remove this empty if-block. I suggest to use the greater than operator for comparing so that you can start with the highest value in the first if-statement and go down to the lowest value in the else-block.

Vocaluxe/Screens/CScreenScore.cs
```diff
@@ -250,6 +253,40 @@ private void _BuildProgressBarString(ref List<string> progressBars)
...
+            {
+                // No sound is played
+            }
+            else if (maxPoints < 5000)
+            {
+                await Task.Delay(500);
+                _ApplauseStream = PlaySound(ESounds.ApplauseLow, 80);
```
flokuep: We should think about the volume level. If I remember correctly, it should be the             CConfig.BackgroundMusicVolume level on that screen. So you could try if it's enough to set it to this value, add some points to it to be louder than that. Or even add another volume to the configuration for sound effects.

flokuep (changes requested):
Wow! Cool changes 🎉 I've found one little code smell part and added a suggestion regarding the volume level of the sound effects to the review.

Vocaluxe/Screens/CScreenScore.cs
```diff
@@ -250,6 +253,40 @@ private void _BuildProgressBarString(ref List<string> progressBars)
...
+        private async void _PlayApplauseSound(int maxPoints)
+        {
+            // Play the appropriate applause sound based on maxPoints
+            if (maxPoints < 2000)
+            {
+                // No sound is played
+            }
+            else if (maxPoints < 5000)
+            {
+                await Task.Delay(500);
```
flokuep: I wonder how the delay is working here with blocking the hole code. Are you sure that this is the right way (to be honest: haven't used C# for years, so I might just missed something).

PR #730, opened 2024-11-20
fix: adjust rendering box for large characters

### LINKED ISSUE #728: Text overlapping on highscore screen
A user reported on discord that with the latest nightly the scores are overlapping on the highscore screen:

This only happens with medleys, might be introduced with #714.

flokuep:
Trying to fix #728. My idea is the following:
- Characters are rendered not directly but for every character (including digits) is rendered in it's max size and stored inside a texture list. These textures (Glyphs) are used for displaying texts.
- During medleys a countdown with big numbers is shown
- For rendering of the glyphs, we use for technically reasons a 1000x1000 box
- this box is (hopefully) too small after increasing the rendering size
- After singing a medley the digit glyphs are someway corrupt

I've increased the box to fix that hopefully.


flokuep:
@marwin89 @Rap-Voca I'm still out of a windows system, so could someone of you test this pr-build: https://ci.appveyor.com/api/buildjobs/mmcmcgiwgx80ycl1/artifacts/Vocaluxe_flokuep-patch-1_Windows_x64.zip

ghost:
It works for me. Thank you!

flokuep:
Also reported as fixed on discord. Thanks for testing 🙃

PR #746, opened 2025-01-05
Fix tone helper drawn on wrong octave

Jiiniasu:
This change addresses the behaviour of which the tone helper is drawn on the wrong octave when singing a line containing notes that span past the first octave from the baseline note. The faulty behaviour was noted in the demo for #704.

The tone helper will be clamped to within one octave of the target note.


flokuep (approved):
Looks good, thanks!

PR #751, opened 2025-01-13
Add a back button in the main options screen.

JohnBergago:
To be consistent with all other screens (except sing screen), we (@DirkHeinke and me) added the Back-Button to the main options screen.

grafik


flokuep (changes requested):
Thanks for your contribution!

For all screen changes, we have a versioning system in place, so when you add or remove elements to/from a screen bump its screen version and add an entry to the according change log.

JohnBergago:
I updated the version and added a changelog entry. I did reorder the existing changelog, as it was different from the other changelog files.

flokuep (approved):
Thanks!

PR #762, opened 2025-03-24
Option to skip Player Selection Screen

Edi316:
A new option has been added to the song selection screen, allowing players to bypass the player selection screen, as it’s often unnecessary to choose players before each song.

SongOptions


flokuep:
Thanks! Have you tested that when playing a duet? There is some code in the name selection screen according to set players to duet parts and I'm not sure what happens if we skip that...

Edi316:
When skipping the selection screen for a duet song, all players will automatically be assigned to Player 1’s duet part.

Edi316:
> Thanks! Have you tested that when playing a duet? There is some code in the name selection screen according to set players to duet parts and I'm not sure what happens if we skip that...

When skipping the selection screen for a duet song, all players will automatically be assigned to Player 1’s duet part.

flokuep:
Okay, thanks! That could be improved in a follow up PR if necessary. 

Thanks for implementing that 🙃

flokuep (approved):

PR #772, opened 2025-04-20
Elevating sound for Progress Bar on Score screen

Edi316:
This introduces an elevating sound for the Progress Bar on the Score screen

The new sound plays first, followed by the applause sound. Additionally, the applause audio files have been updated to minimize the initial silence to make it more seamless.

See #675:
"On rating screen: Elevating Sound to illustrate pointsbar progress (5seconds =100%, fade out sound earlier according to score)"


flokuep (approved):

flokuep:
Thanks! As there are now a lot of effects right now, there might be the need of adjusting there volume. If you like, you could think about adding a sound effects volume configuration (which also provides the possibility to turn that off completely.)

Issue #682, opened 2024-02-06
Re-design option screen


marwin89:
### What
- Re-design option screen.
- Organizing options via tabs

### Why
- we need more space for configurations (e.g. smartphone, dev-options, accessibility)
- right now there is not enough space for more categories in option screen.
- give players better orientation through icons

### Mock up
Here is a mockup. It maybe not perfect but it's sufficient for programmers to implement. A UI expert can finetune afterwards. 

![mockup-optionscreen](https://github.com/Vocaluxe/Vocaluxe/assets/63517874/1761c48c-a764-41a2-a0b6-3b2806a009e4)

PR #775, opened 2025-04-25
Graphics Options

Edi316:
This introduces the following Graphics Options to the Options screen, which were previously only accessible via the configuration file:
Texture Quality
Cover Size
Fullscreen
Stretched

Due to limited space in the current Options screen, these settings have been added to the Lyrics Options, which have now been renamed to “Lyrics/Graphics.” 
In the future, these graphics options can easily be moved to a dedicated screen during the redesign of the Options screen (see  #682).

### DIFF
--- a/Vocaluxe/Screens/CScreenOptionsLyrics.cs
+++ b/Vocaluxe/Screens/CScreenOptionsLyrics.cs
@@ -15,6 +15,7 @@
 // along with Vocaluxe. If not, see <http://www.gnu.org/licenses/>.
 #endregion
 
+using System;
 using System.Windows.Forms;
 using System.IO;
 using Vocaluxe.Base;
@@ -164,7 +165,6 @@ namespace Vocaluxe.Screens
                         
             CConfig.Config.Graphics.FullScreen = (EOffOn)_SelectSlides[_SelectSlideFullScreen].Selection;
             CConfig.Config.Graphics.Stretch = (EOffOn)_SelectSlides[_SelectSlideStretch].Selection;
-            CConfig.Config.Graphics.VSync = (EOffOn)_SelectSlides[_SelectSlideVSync].Selection;
             
             CConfig.SaveConfig();
         }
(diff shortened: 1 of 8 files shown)


Vocaluxe/Screens/CScreenOptionsLyrics.cs
```diff
@@ -40,14 +47,31 @@ public override void Init()
...
+            int index = Array.IndexOf(options, currentCoverSize.ToString());
+            _SelectSlides[_SelectSlideCoverSize].Selection = index;
+            
+            _SelectSlides[_SelectSlideFullScreen].SetValues<EOffOn>((int)CConfig.Config.Graphics.FullScreen);
+            _SelectSlides[_SelectSlideFullScreen].Selection = (int)CConfig.Config.Graphics.FullScreen;
+
+            _SelectSlides[_SelectSlideStretch].SetValues<EOffOn>((int)CConfig.Config.Graphics.Stretch);
+            _SelectSlides[_SelectSlideStretch].Selection = (int)CConfig.Config.Graphics.Stretch;
+
+            _Texts[_TextWarningRestart].Visible = true;
```
flokuep: What about only showing that warning if there is a change that requires a restart?

Vocaluxe/Screens/CScreenOptionsLyrics.cs
```diff
@@ -40,14 +47,31 @@ public override void Init()
...
 
         public override void LoadTheme(string xmlPath)
         {
             base.LoadTheme(xmlPath);
             _SelectSlides[_SelectSlideLyricStyle].SetValues<ELyricStyle>((int)CConfig.Config.Theme.LyricStyle);
             _SelectSlides[_SelectSlideLyricsPosition].SetValues<ELyricsPosition>((int)CConfig.Config.Game.LyricsPosition);
+
+            _SelectSlides[_SelectSlideTextureQuality].SetValues<ETextureQuality>((int)CConfig.Config.Graphics.TextureQuality);
+            
+            _SelectSlides[_SelectSlideCoverSize].AddValues(new string[] { "32", "64", "128", "256", "512", "1024" });
```
flokuep: You could move the string array to a constant as you need it several times. You could also check the settings class if there is already a constant including these values.

flokuep (changes requested):
Thanks for working on improving Vocaluxe 🙃 I've found two small things that could be improved. Overall your code looks fine.

flokuep (approved):

PR #776, opened 2025-04-28
Add Song Folder in Game Options

Edi316:
This adds a button to the Game Options menu that allows you to add a song folder. Previously, this could only be done by editing the configuration file.

### DIFF
--- a/Output/Languages/en.xml
+++ b/Output/Languages/en.xml
@@ -206,10 +206,12 @@
   <string name="TR_SCREENO_BUTTON_SOUND">Sound</string>
   <string name="TR_SCREENO_BUTTON_RECORD">Record</string>
   <string name="TR_SCREENO_BUTTON_VIDEO">Video</string>
-  <string name="TR_SCREENO_BUTTON_LYRICS">Lyrics</string>
+  <string name="TR_SCREENO_BUTTON_LYRICS">Lyrics/Graphics</string>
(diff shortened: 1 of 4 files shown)


flokuep:
Thanks for working on that! I'm a little bit afraid of merging this feature in its current state:
- it's unclear to the user that the current configuration is completely overwritten
- it does not support multiple song folders
- the restart info is shown even if the user cancels the folder dialog

The most complete implementation could come with a new screen (or a popup screen, should be both doable) that shows a list of the currently configured song folders, provides the ability to remove entries from the list and add new ones to it. This screen needs to ensure that at least one song folder is configured. 

A smaller implementation could follow your way but maybe just add the folder to the list of song folders?

As a follow up you could try to restart the song loading after changing this configuration. But I'm not sure what needs to be done for that, but it should be working somehow.

Edi316:
Thank you for the feedback

- When a folder is selected, it will be added as an additional folder in the configuration. No existing folders will be deleted or overwritten. I have included screenshots to show the current behavior.

- I have updated the restart prompt so that it only appears when a folder is selected. Pressing “Cancel” will no longer trigger the restart prompt.

TestCase

flokuep:
Ah, sounds good. I've just missed that you used the existing list for the new configuration. Feel free to open the pr again for merging.

flokuep (approved):

Issue #627, opened 2023-11-19
Introduce #VOCALS


marwin89:
## Suggestion:

- Introduce #VOCALS Attribute in Song parser/ txt-file.
- This Attribute points to audio file with only vocals (accapella)
- Feature: volume of vocals can be adjusted (while #INSTRUMENTAL playing n sync)
- Feature: Accapella mode select in song screen
- Normal mode plays # AUDIO (music vocals in one file combined)
- search for files matching the following scheme in case of the tags are missing:

`#VOCALS: Artist - Title of the song [VOC].mp3`

`#INSTRUMENTAL: Artist - Title of the song [INSTR].mp3`

(according to [official common Song format Specification](https://usdx.eu/format/) v1.1.0 - release on: 1th Dec 2023)

flokuep:
That's the most challenging change for the new song format, as it has impact on the game logic. I would split that into several parts, next to some preparations.

Preparations:
- Adding new tags with new files is easy, can be done in the same way as it is currently done for the MP3 tag.

Step 1: Playing multiple files in sync
- For that feature we need to have multiple audio files played in sync, most of the logic for the song played during singing is handled within CSCreenSing.
- Have a look on the _CurrentStream integer, as this holds an identifier for the currently played audio file.
- I suggest to do a proove-of-concept here first by duplication that _CurrentStream stuff and check if it is possible to play multiple files at the same time (with a random audio file) and if you can keep both streams in sync (with the same audio file)
- If that is sucessfull, you can extend the code to handle all three audio streams - Otherwise, a rewrite of the audio stuff is necessary (have a look on the background-music stuff, maybe that's helpful).

Step 2: Add game logic for it
- As a first simple step I suggest to implement switching through the different modes by pressing a key

Step 3: UI enhancements
- Draw some concepts how these features could be integrated into the ui. There is a song menu where another selection slider can be added, but if I remember correctly, that menu is already compex
- The volume adjustment could be added to the pause screen for example.

Complexity: L to XL.

bigboss97:
Love to see this feature 😊
I guess "multiple audio files played in sync" is a big job. But if the channels are mixed in real time the user can even adjust the level of the vocal separately 👍

[...]

PR #788, opened 2025-05-08
Gameplay support for #VOCALS

Edi316:
This introduces gameplay support for the vocals tag. (#627)

When a song includes both Instrumental and Vocal tags, a new Audio Mode option, "Instrumental/Vocal", becomes selectable. 
In this mode, the volume of the vocal track can be adjusted from the pause screen.

The new Audio Mode and volume slider are only visible when a song has both tags and the Instrumental/Vocal mode is selected. Otherwise, these controls remain hidden.

Additionally, headings have been added to the song options, making it easier for users to understand the different slider settings.

### DIFF
--- a/Vocaluxe/Screens/CScreenSing.cs
+++ b/Vocaluxe/Screens/CScreenSing.cs
@@ -380,15 +380,11 @@ namespace Vocaluxe.Screens
                             float newTime = _CurrentTime + (keyEvent.Mod == EModifier.Shift ? 10f : 30f);
                             if (CSound.GetLength(_CurrentStream) < newTime)
                                 newTime = CSound.GetLength(_CurrentStream) - 1f;
-                            
+                                CSound.SetPosition(_CurrentStream, newTime);
+                                
                             if (CScreenSong.GetAudioMode() == EAudioMode.TR_AUDIOMODE_VOCALS)
                             {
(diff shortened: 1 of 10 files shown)


Vocaluxe/Screens/CScreenSing.cs
```diff
@@ -364,7 +380,16 @@ public override bool HandleInput(SKeyEvent keyEvent)
...
+                            
+                            if (CScreenSong.GetAudioMode() == EAudioMode.TR_AUDIOMODE_VOCALS)
+                            {
+                                CSound.SetPosition(_CurrentStream, newTime);
+                                CSound.SetPosition(_CurrentStreamVocals, newTime);
+                            }
+                            else
+                            {
                             CSound.SetPosition(_CurrentStream, newTime);
+                            }
```
flokuep: As the position of _CurrentStream needs to be set in both cases you can remove the else so that it will be set in every case.

flokuep (approved):
Thanks for implementing that! Found just one little code smell that could be improved. 

I haven't thought that it would be so easy to keep the streams in sync, glad to see it.

flokuep (approved):

marwin89:
@Edi316 , cool! thanks a lot for your contribution 🚀 👍🏻 👍🏻

PR #790, opened 2025-05-19
Option screen re-design

Edi316:
This re-design updates the options screen as described in issue #682. 
The new layout provides additional space, making it easier to add more options in the future.

The icons currently in use are sourced from uxwing.com (free to use icons with no attribution required). These icons can be easily replaced by a UI artist/designer with alternative/better designs if necessary.

Screen1

Screen2


flokuep (changes requested):
Thanks for working on that! I really like the overall idea for the config screen! 

I'm not sure if I like the icon style and padding. But I'm not sure if I find enough time to look for some others...

Within the code, I've seen a lot of duplications. Have you thought about adding a new super class between the CScreenOptionsX and the CMenu? Eg. a CAbstractScreenOptions that includes all the relevant stuff for the options navigation? With that change you should be able to reduce the amount of duplications dramatically.

Edi316:
Thank you for your feedback. 
I will close this commit for now and resubmit it once I have implemented the changes.

Issue #698, opened 2024-04-19
Migrate to GitHub Actions


flokuep:
To reduce some dependencies and have a more smooth integration into GitHub, we should consider switching to GitHub Actions for building nightly and release versions. 

The most complicated part of achieving this is the version number handling which is solved by some scripts currently. 

While doing this, we could think about skipping transifex-pullrequests automatically.

flokuep:
@Jiiniasu I've seen that you already made some experience with that topic, on your forks. Any knowledge or solved problems you can share with us?

Jiiniasu:
The Github actions I have in my SSIS-Remote branch is my first experience with it, so can't say I have too much knowledge/learnings that aren't basic usage. 

How is the current Appveyor set up? Are there parts that can only be seen with some kind of login to the service?

flokuep:
Ah, okay. Same here :) I've just playing around with GH Action on my fork...

No, all Appveyor configuration should be within the appveyor.yaml and the referenced build-scripts within the repository itself.

poupounetjoyeux:
Old subject but following my PR build issue, I have some knowledge on github actions for building /releasing
Since it will be smoother with a recent .NET version I will firstly try to upgrade it then if I have some time will check to write needed workflows 👌

flokuep:
There is already an open pr for it, you could check. It still on my to-do list to find some time to test it...

PR #793, opened 2025-06-23
GitHub action migration

JohnBergago:
We (@DirkHeinke and me) started working on #698.

For now we created only the pipeline for nightly builds. We introduced a new version numbering for nightly builds based on the build date. Basically it is the [same scheme Microsoft uses](https://stackoverflow.com/questions/17720008/what-causes-assembly-version-incrementation-when-using-asterisk#:~:text=The%20build%20number%20is%20the%20number%20of%20days%20since%20Jan%201st%2C%202000%2C%20the%20revision%20number%20is%20the%20number%20of%20seconds%20since%20midnight%20(no%20DST%20correction)%2C%20divided%20by%20two) when version is set to `0.0.*`.  

We decided to not use the current version number as prefix, to distinguish nightly builds from release builds. As a consequence nightly versions will always have a smaller version tag compared to releases, as the latter are the preferred stable version. 

We will continue working on this PR and setup a pipeline for release builds. But we'd like to get early feedback from you.


flokuep:
Great, thank you! I've tried that on my own some month ago, but never had the chance to finish it. You could have a look on it if you run across problems, maybe I've already solved them 🙃 https://github.com/flokuep/Vocaluxe/tree/develop/.github

JohnBergago:
We implemented the whole apveyor pipeline in Github Actions. How to proceed with this PR?

The following things were changed or implemented:

## Workflows

### nightly-build.yaml

- runs on pushes to `develop`
- runs the build.yaml workflow with tests
- creates a Nightly version release
- the resulting version string is `0.0.[days_since_2000-01-01].[seconds_since_midnight/2]`
- This ensures an automatically increasing nightly build version

### release-build.yaml

- runs on pushes of tags in the format of `[0-9]+.[0-9]+.[0-9]+`
- runs the build.yaml workflow with tests
- creates a release with the specified tag as version string

### pr-build.yaml
- runs on pull requests
- runs the build.yaml workflow with tests
- does not create a release

### build.yaml

- used by nightly and release build workflows
- Runs a parallel build of `x86` and `x64` with the following steps:
  - Setup build enviroment and restore nuget packages
  - build the solution with specified version string
  - includes theme videos
  - runs tests
  - uploads test results and build artifacts
- creates a release depending on the specified tag

### sonarscanner.yaml

- Pretty much the default sonar suggestion
- runs on develop and pull requests
- **Needs environment setup on github:**
  -  `SONAR_TOKEN` must be set in `Project Settings > Secrets and Variables > Actions > Repository Secrets`

### test-reporter.yaml
- Uploads the test results once the pipeline ran
- However, according to the [docs](https://github.com/dorny/test-reporter/tree/b082adf0eced0765477756c2a610396589b8c637/) (and our own tests) it needs to be part of the default branch to work:
  > The second workflow will only run after it has been merged into your default branch (typically main or master), it won't run in a PR unless after the workflow file is part of that branch.

## State of the Tests 

You may notice, that the pipeline fails right now. The reason seems to be changes made by PR #823. It also fails in the appveyor build, but nobody ever noticed.

Additionally we skipped some of the parser tests, since they also failed due to changes in the base file.

McSchneidar:
@JohnBergago
Could you please update the branch to check if the latest changes resolve (some of) the issues you encountered.
Thank you!

bohning:
The error occurs because the SONAR_TOKEN secret is not being passed to the workflow when a pull request is opened or synchronized, resulting in an empty value.  This is a common security restriction in GitHub Actions, especially for workflows triggered by external contributors or automated bots like Dependabot. 

The primary solution is to conditionally skip the SonarQube analysis for pull requests and only run it on direct push events to the develop branch.  This is the safest and most straightforward approach.

Update your workflow by adding an if condition to the Build and analyze job:
```
- name: Build and analyze
        if: github.event_name == 'push'
        env:
          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
        shell: powershell
        run: |
          ${{ runner.temp }}\scanner\dotnet-sonarscanner begin /k:"Vocaluxe" /o:"vocaluxe" /d:sonar.token=$Env:SONAR_TOKEN
          msbuild Vocaluxe.sln /p:Configuration=ReleaseWin /p:Platform="x64"
          ${{ runner.temp }}\scanner\dotnet-sonarscanner end /d:sonar.token=$Env:SONAR_TOKEN
```
This ensures the SONAR_TOKEN is only used in push events where the secret is available, preventing the "invalid format" error during PRs. Code quality analysis will still run on every commit to develop, which is usually sufficient.

JohnBergago:
> The error occurs because the SONAR_TOKEN secret is not being passed to the workflow when a pull request is opened or synchronized, resulting in an empty value. This is a common security restriction in GitHub Actions, especially for workflows triggered by external contributors or automated bots like Dependabot.
> 
> The primary solution is to conditionally skip the SonarQube analysis for pull requests and only run it on direct push events to the develop branch. This is the safest and most straightforward approach.
> 
> Update your workflow by adding an if condition to the Build and analyze job:
> 
> ```
> - name: Build and analyze
>         if: github.event_name == 'push'
>         env:
>           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
>         shell: powershell
>         run: |
>           ${{ runner.temp }}\scanner\dotnet-sonarscanner begin /k:"Vocaluxe" /o:"vocaluxe" /d:sonar.token=$Env:SONAR_TOKEN
>           msbuild Vocaluxe.sln /p:Configuration=ReleaseWin /p:Platform="x64"
>           ${{ runner.temp }}\scanner\dotnet-sonarscanner end /d:sonar.token=$Env:SONAR_TOKEN
> ```
> 
> This ensures the SONAR_TOKEN is only used in push events where the secret is available, preventing the "invalid format" error during PRs. Code quality analysis will still run on every commit to develop, which is usually sufficient.

Thanks for pointing that out. To us there are two possible solutions:

1. The one @bohning proposed, which would result in SonarQube not checking pull requests.
2. Trying to use [SonarQubes automatic analysis](https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/automatic-analysis). Someone with SonarQube access needs to configure it.

Which one do you prefer?

PR #804, opened 2025-09-02
Bumped max count of filtered songs to 800

kmkacpermajor:
I have 650 polish songs, currently list just cuts after 500th.

So I bumped max count of filtered songs to 800.
Runs fine with over 2500 songs on my 6GB RAM laptop.

It has already been bumped from 200 to 500 by Stefan1200-de 5 years ago.


flokuep (approved):
Thanks!

PR #811, opened 2025-11-03
Adding a regex based solution for enumerating files looking for multiple extensions at same time

iakona:
This solution was based on this stack overflow answer https://stackoverflow.com/a/3754470

Based on this comment, you may be able to get better perf but doing multiple parallel `Directory.EnumerateFiles` calls, but then the result ordering isn't consistent anymore, and I'm not sure if that would be an issue for this program or not.

### DIFF
--- a/VocaluxeLib/CHelper.cs
+++ b/VocaluxeLib/CHelper.cs
@@ -28,9 +28,9 @@ namespace VocaluxeLib
 {
     public static class CHelper
     {
-        private static readonly string _TextFileTypesRegex = "\\.txt|\\.txd";
-        private static readonly string _SoundFileTypesRegex = "\\.mp3|\\.wma|\\.ogg|\\.wav";
-        private static readonly string _ImageFileTypesRegex = "\\.jpg|\\.jpeg|\\.png|\\.gif";
+        private static readonly Regex _TextFileTypesRegex = new Regex("\\.txt|\\.txd", RegexOptions.IgnoreCase);
+        private static readonly Regex _SoundFileTypesRegex = new Regex("\\.mp3|\\.wma|\\.ogg|\\.wav", RegexOptions.IgnoreCase);
+        private static readonly Regex _ImageFileTypesRegex = new Regex("\\.jpg|\\.jpeg|\\.png|\\.gif", RegexOptions.IgnoreCase);
 
         public static int CombinationCount(int n, int k)
         {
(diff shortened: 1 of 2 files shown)


iakona:
my friend has a setup with around 6k songs, and change in `LoadSongs()` sped up loading time for him by around 20%

VocaluxeLib/CHelper.cs
```diff
@@ -192,16 +181,45 @@ public static IEnumerable<string> ListFiles(string path, string searchPattern, b
...
+            var dir = new DirectoryInfo(path);
+            if (!dir.Exists)
+                return files;
+
+            if (recursive && !fullpath)
+                throw new NotSupportedException("recursive file listing with relative path names is not supported");
+
+            try
+            {
+                Regex extPatternRegex = new Regex(extPattern, RegexOptions.IgnoreCase, TimeSpan.FromSeconds(1));
```
iakona: quality gate told me to add a default timeout, I put in 1 second, but not I'm too sure what would be reasonable here

flokuep: Not sure about it. Should be fine to use that without a default timeout.

VocaluxeLib/CHelper.cs
```diff
@@ -192,16 +181,45 @@ public static IEnumerable<string> ListFiles(string path, string searchPattern, b
...
+                files = Directory.EnumerateFiles(path, "*", recursive ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly)
+                    .Where(file => extPatternRegex.IsMatch(Path.GetExtension(file)));
+                files = files.Select(fullpath ? (Func<string, string>)Path.GetFullPath : Path.GetFileName);
+            }
+            catch (Exception) { }
```
flokuep: Please catch expected exceptions here and forward them into our log. You should find many examples within the code for that.

flokuep (changes requested):
Thanks for improving this :) I'd like to see the error handling improved. Everything else looks fine!

VocaluxeLib/CHelper.cs
iakona: alright, i'll switch it back

VocaluxeLib/CHelper.cs
iakona: yeah should be easy enough, i was just copy pasting the other list files function as a base, but i'll clean them both up

VocaluxeLib/CHelper.cs
```diff
@@ -21,27 +21,16 @@
...
-            };
-
-        private static readonly List<string> _ImageFileTypes = new List<string>
-            {
-                "*.jpg",
-                "*.jpeg",
-                "*.png",
-                "*.gif"
-            };
+        private static readonly Regex _TextFileTypesRegex = new Regex("\\.txt|\\.txd", RegexOptions.IgnoreCase);
```
iakona, line 31: i'll just store the regex directly, so we can save on constructing the object each execution

VocaluxeLib/CHelper.cs
```diff
@@ -186,22 +175,58 @@ public static IEnumerable<string> ListFiles(string path, string searchPattern, b
                 files = Directory.EnumerateFiles(dir.FullName, searchPattern, recursive ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly);
                 files = files.Select(fullpath ? (Func<string, string>)Path.GetFullPath : Path.GetFileName);
             }
-            catch (Exception) {}
+            catch (Exception e)
+            {
+                CLog.Error(e, "Error listing files for path " + path + " with search pattern " + searchPattern);
+                return Enumerable.Empty<string>();
```
iakona, line 181: not sure if you'd rather keep the return the same as it was before where you potentially have a mis-formated list if the `Select` call throws the exception instead of `EnumerateFiles` throwing it

flokuep (approved):

PR #813, opened 2025-11-20
Adding new config option to load the songs in the background (similar to covers) to allow the user to start using the program earlier

iakona:


Vocaluxe/Screens/CScreenLoad.cs
```diff
@@ -140,9 +140,10 @@ public override bool UpdateGame()
...
 
-            bool next = CConfig.Config.Theme.CoverLoading != ECoverLoading.TR_CONFIG_COVERLOADING_ATSTART || CSongs.CoverLoaded;
+            bool songsReady = CConfig.Config.Theme.SongLoading != ESongLoading.TR_CONFIG_SONGLOADING_ATSTART || CSongs.SongsLoaded;
+            bool coversReady = CConfig.Config.Theme.CoverLoading != ECoverLoading.TR_CONFIG_COVERLOADING_ATSTART || CSongs.CoverLoaded;
 
-            if ((_IntroOutPlayed || _SkipIntro) && next && CSettings.ProgramState == EProgramState.Start && CSongs.SongsLoaded)
+            if ((_IntroOutPlayed || _SkipIntro) && CSettings.ProgramState == EProgramState.Start && songsReady && coversReady)
             {
                 CSettings.ProgramState = EProgramState.Normal;
                 CGraphics.FadeTo(EScreen.Main);
```
iakona, line 149: the `FadeTo` call will cause the song loading count text to disappear and then reappear when dynamic is enabled. not sure if it's necessarily worth going down the rabbit hole to try to fix

Vocaluxe/Base/CCover.cs
```diff
@@ -112,7 +112,7 @@ public static CTextureRef GenerateCover(string text, ECoverGeneratorType type, C
                 }, _CancelToken.Token);
             lock (_Covers)
             {
-                _Covers.Add(text, texture);
+                _Covers[text] = texture;
```
iakona, line 115: i encountered a crash when closing sometimes where we'd try to add a cover that already existed, so switching this to an upsert operation fixed that

Vocaluxe/Base/CSongFilter.cs
```diff
@@ -66,6 +66,11 @@ public EDuetOptions DuetOptions
             }
         }
 
+        public void ForceRefresh()
+        {
+            _SetChanged();
```
iakona, line 71: this will cause the visible song list to update

Vocaluxe/Base/CSongFilter.cs
```diff
@@ -180,8 +185,9 @@ private void _FilterSongs()
                 }
             }
 
-            foreach (CSong song in CSongs.Songs)
+            for (int i = 0; i < CSongs.NumAllSongs; i++)
             {
+                CSong song = CSongs.Songs[i];
```
iakona, line 190: i encountered a modification of enumerable while enumerating, so changing this to be index based fixed that

Vocaluxe/Screens/CScreenMain.cs
```diff
@@ -148,6 +152,22 @@ public override bool UpdateGame()
             bool profileOK = CProfiles.NumProfiles > 0;
             _Statics[_StaticWarningProfiles].Visible = !profileOK;
             _Texts[_TextWarningProfiles].Visible = !profileOK;
+            if (CConfig.Config.Theme.SongLoading == ESongLoading.TR_CONFIG_SONGLOADING_DYNAMIC)
```
iakona, line 155: when dynamic is enabled, figured it'd be a good idea to still show the user how they are progressing towards loading songs. but i can always remove this if you'd like

Vocaluxe/Screens/CScreenSong.cs
```diff
@@ -202,6 +203,10 @@ public override void Init()
             _TimerShortInfoText = new System.Timers.Timer(5000);
             _TimerShortInfoText.AutoReset = false;
             _TimerShortInfoText.Elapsed += OnTimedEventShortInfoText;
+
+            _TimerLoadSongs = new System.Timers.Timer(5000);
```
iakona, line 207: i used 5 seconds as a value since other timers (like above) run on that interval. i'm open to suggestions to other refresh rates

Vocaluxe/Base/CConfig.cs
```diff
@@ -125,6 +125,7 @@ public struct SConfigTheme
             [DefaultValue(EPlayerInfo.TR_CONFIG_PLAYERINFO_BOTH)] public EPlayerInfo PlayerInfo;
             [DefaultValue(EFadePlayerInfo.TR_CONFIG_FADEPLAYERINFO_OFF)] public EFadePlayerInfo FadePlayerInfo;
             [DefaultValue(ECoverLoading.TR_CONFIG_COVERLOADING_DYNAMIC)] public ECoverLoading CoverLoading;
+            [DefaultValue(ESongLoading.TR_CONFIG_SONGLOADING_ATSTART)] public ESongLoading SongLoading;
```
iakona, line 128: defaulting this to do the song loading upfront to preserve existing functionality, but this also means there's less eyes on the feature. not sure if it's better to enable this feature by default in nightly/prerelease and then default it to up front on the stable release (whenever this feature goes out)

iakona (commented):
i've done testing on my end and with a friend and this generally seems to work. but there may still be some edge cases i haven't handled properly yet. in theory though rolling this out with nightly/prerelease would get more eyes to help fix an edge cases i have created

Output/Languages/en.xml
```diff
@@ -63,6 +63,8 @@
   <string name="TR_CONFIG_COVERLOADING_ONDEMAND">On demand</string>
   <string name="TR_CONFIG_COVERLOADING_ATSTART">At startup</string>
   <string name="TR_CONFIG_COVERLOADING_DYNAMIC">Dynamic</string>
+  <string name="TR_CONFIG_SONGLOADING_ATSTART">At startup</string>
```
iakona, line 66: i should probably just use the coverloading strings here instead right? since they are already translated? or should i copy paste those translations into the other languages?

flokuep:
Thank you for your contribution! I'm not sure what you're trying to achieve with this pull request? Cover loading on demand is used to only load cover into the memory if they are needed, so safe some memory as pictured tend to be bigger.

Song information are in contrast to that necessary to display the list of songs, so lazy loading wouldn't be that helpful? Plus song information aren't that big.

iakona:
> Thank you for your contribution! I'm not sure what you're trying to achieve with this pull request? Cover loading on demand is used to only load cover into the memory if they are needed, so safe some memory as pictured tend to be bigger.
> 
> Song information are in contrast to that necessary to display the list of songs, so lazy loading wouldn't be that helpful? Plus song information aren't that big.

what i'm trying to accomplish is getting into the singing faster. currently all songs must be loaded before you're able to interact with anything on the main menu, and for folks with large song libraries and/or on low spec computers that could be 5+ minutes of waiting. so the idea is to have them be able to jump into using the program, setting up the microphone and singers, tweaking settings, or even looking at the song list some while things load in the background. instead of having to wait a period of time first before doing things

flokuep:
Got it! As you've already stated in you description: there could be a lot of edge cases happening and stuff will become quite complex. 

Have you tried to identify bottlenecks of the song loading? Maybe there is room for improvement? If I remember correctly, you can activate a benchmark within the config and check the benchmark log afterwards.

iakona:
> Got it! As you've already stated in you description: there could be a lot of edge cases happening and stuff will become quite complex.
> 
> Have you tried to identify bottlenecks of the song loading? Maybe there is room for improvement? If I remember correctly, you can activate a benchmark within the config and check the benchmark log afterwards.

my local setup isn't really in a position to benchmark, i've only got a pretty small library and a nice computer. my friend has a large library with a not so nice computer, but also doesn't have development environment setup. is there a way to enable the benchmark on a build i can send them and then they send me back the data?

flokuep:
Cannot remember where it's located. Could you check the Vocaluxe.log for benchmark information? Or/and set             DebugLevel to TR_CONFIG_ON within config.xml?

PR #823, opened 2025-12-29
Modernize NuGet dependencies and migrate to PackageReference

metinc:
This PR updates and modernizes the NuGet dependency setup.

Changes included:
- Migrate NuGet usage to PackageReference
- Replace vocaluxe.dependencies.sqlite with Microsoft.Data.Sqlite
- Update Newtonsoft.Json and Serilog package references to the latest versions

Why:
- Reduce restore/build friction by using the current .NET-standard PackageReference approach
- Move off a custom/legacy SQLite dependency onto the supported Microsoft.Data.Sqlite provider
- Microsoft.Data.Sqlite should make Linux support simpler (more standard cross-platform provider and packaging)
- Keep core dependencies up to date for stability, compatibility, and maintenance

Built and smoke-tested in VS Code and Visual Studio 2026.

Let me know if I should change anything.


flokuep:
That's great, thank you! Do you expect compability problems between sqlite files written by the old version and files from the new version?

flokuep (approved):

metinc:
No, there are no breaking changes. I tested it to make sure the user can keep the old sqlite files.

PR #832, opened 2026-01-24
Two small fixes

ghost:
This addresses two rare crash scenarios I encountered:

**CNameSelection**: 
- Added check for  _VisibleProfiles.Count > 0  before setting  SelectedID  and  ActualSelection .
- Leaves selection empty when no profiles available, avoiding implicit first-entry access and potential IndexOutOfRangeException.

	
**CPlaybackBase**: 
- Wrapped stream iteration in  lock (_Streams)  to prevent race conditions from concurrent access.
- Ensures safe volume adjustments during playback without deadlocks in multi-threaded scenarios.


flokuep (approved):
Thanks!

PR #846, opened 2026-03-19
Fix for 2 build errors (non-blocking)

McSchneidar:
### PartyModeChallenge
Constructor takes `numRounds` as upper limit, but dynamically generates more rounds until all players have sung equally often.  
Test expected exactly `numRounds`, but got "fair" amount.  
Test adapted to match actual logic.

### Serilog
NuGet restore installed Serilog 4.2.0, but manifest conflict (HRESULT 0x80131040).  
Added binding redirects


flokuep (approved):
Thanks! 🎉

Issue #684, opened 2024-02-06
Add grid overview to profile screen


marwin89:
### What
Give the profile screen a sidebar with an tile overview of profiles.
By clicking on a tile: Profile is automatically selected

### Why
It makes it easer for players to browse large profile collection.

### Mock up
![ProfileScreenWithSelectSidebar](https://github.com/Vocaluxe/Vocaluxe/assets/63517874/9a13e4ed-1c20-422c-a212-c14797bdb301)

### Additional Info
There has to be only some small coding. Just copy the feature from Playername screen. Bigger part is the layouting.

Edi316:
Issue can be closed (resolved by #858)

PR #858, opened 2026-04-01
Grid overview for profile screen

Edi316:
This adds a grid overview to the profile screen as described in #684


flokuep:
Would you add a screenshot of your changes?

Edi316:
Here is the screenshot

Screen

flokuep (approved):

Issue #768, opened 2025-04-11
Upgrade .NET to a modern cross platform version


dr-waterstorm:
As suggested by @aharter in the #673 we should upgrade .NET from the old "legacy" (but still supported) 4.x version to a more modern and cross platform version of .NET. Probably .NET 8 (the current LTS) would be the best choice at the moment.

This would give us a few benefits. 
- It would be easier to port / compile this project on Linux and MacOS as .NET 8 supports both platforms. 
- It would remove the requirement for the .net runtime
- It's faster and more modern (at least MS states this, I've never migrated a project myself before)

The discussion here https://github.com/Vocaluxe/Vocaluxe/issues/673#issuecomment-2564766909 gives more insight into the topic.

flokuep:
@Edi316 Would be happy to see that issue resolved - If you have time for that and need a new task, that would be a good one 🙃

PR #870, opened 2026-04-26
Linux support via nix dev shell

Niketin:
Vocaluxe can now be built and run on Linux using Mono and a nix dev shell.

## How to build and run

See the updated `HowToBuildLinux.md` for instructions.

## What was needed

Getting this working required fixes across the build system and runtime:

- Nix flake for a reproducible environment with all dependencies
- Mono msbuild doesn't handle NuGet PackageReference out of the box — `Directory.Build.targets` fills that gap
- `strncpy_s` in acinerella.c is MSVC-only, replaced with `strncpy`
- `System.XML` reference casing fails on case-sensitive filesystems
- Mono rejects PNG-compressed `.ico` files — wrapped in try/catch
- GStreamer requires a native glue library not available on Linux — falls back to PortAudio
- FFmpeg 6 decodes MP3 as float planar (FLTP) instead of S16 — acinerella needed swresample + buffer size fixes
- Two Mono 6.x bugs in XML deserialization and profile loading

## Known limitations
- No window icon (cosmetic)
- GStreamer backend unavailable, PortAudio used instead
- Missing default skin assets (pre-existing issue, not Linux-specific)


flokuep:
That's cool! I'd like to merge the GitHub actions stuff in the near future, do you see some conflicts, or something you could adapt/enhance to build some linux binaries?

Niketin:
> That's cool! I'd like to merge the GitHub actions stuff in the near future, do you see some conflicts, or something you could adapt/enhance to build some linux binaries?

Using GitHub Actions sounds like a good direction. Regarding Linux binaries, I think we are currently limited by the .NET Framework; moving to a more modern environment like .NET 10 (as discussed in #768) would likely make the process much smoother.

I initially used Nix to create a reproducible environment, as it was the only way I could get the project to build on Linux successfully. While this PR is a helpful step forward for development, the resulting binaries are tied to the Nix ecosystem. Producing general Linux binaries that run across different distributions without Nix would require significantly more work and infrastructure, so this is likely a starting point rather than a complete solution for universal releases.

PR #881, opened 2026-06-24
Cross-platform port to .NET 10 — Linux + Windows (OpenGL, PortAudio, SkiaSharp) (#768)

byte55:
## TL;DR
Cross-platform port of Vocaluxe to **.NET 10**, addressing issue **#768**. One single codebase that
compiles and runs on **Linux**, and compiles + passes unit tests on **Windows** and **macOS**. The game
runs end-to-end on Linux today (validated by hand) and is now **also validated by hand on Windows** (singing
and party mode played on real hardware) after the fullscreen-present fix below. macOS hands-on validation is
still pending.

- Base: branched from `develop` at `b9b0f7c` (there is no `master`/`main`; `develop` is the trunk).
- 47 commits, 109 files, **+2943 / −4229** (net smaller — several Windows-only backends were dropped).
- All 221 unit tests pass on Linux **and** Windows **and** macOS in CI.

**AI disclaimer:** The port itself was implemented by **Claude Opus 4.8** (Anthropic) working under human
direction. A **human** did the hands-on validation on Linux and Windows (running and playing the game —
singing and party mode — on real hardware); CI covers builds and unit tests across all three OSes. macOS
has **not** been verified by a human yet — please review the code accordingly.

The AI working aids used during the port — the `.claude/` agent definitions and the `CLAUDE.md` project
guide (and this `pr-notes.md` itself) — were deliberately **left out of the commits** so the PR contains
only the code, not AI tooling. They can be contributed separately if the maintainers would find them useful.

## What this branch does
Turns the Windows-only .NET Framework game into a single cross-platform .NET 10 app by replacing every
Windows-specific subsystem with a portable equivalent, behind the existing static-facade/pluggable-backend
architecture (`CDraw`/`CSound`/`CRecord`/`CVideo`/`CWebcam`/…). `VocaluxeLib` still talks to the game only
through `CBase`.

## Migration philosophy: faithful 1:1 port, not a redesign
The aim was a **1:1 migration** — same features, same behavior, same UX as the current release — only on a
portable stack. We deliberately avoided new features, redesigns and opportunistic "while we're here"
optimizations, so the diff stays reviewable and behavior stays comparable to upstream. Windows-only
subsystems with no .NET 10 equivalent were swapped for the portable equivalents listed below, but the
*user-visible behavior* was kept as close as possible.

**The single intentional exception is the startup loading progress bar** (see "Windows runtime bring-up"
below): the original splash showed only a logo, and the multi-second .NET 10 init looked like a hang, so we
added a progress bar + status text. Everything else is meant to match upstream behavior.

## Fundamental changes reviewers/contributors MUST know
These are deliberate and change the **Windows** experience too (it is no longer Direct3D/DirectSound):

- **.NET Framework → .NET 10**, old `.csproj` → **SDK-style** projects. WinForms removed entirely.
- **Rendering: OpenGL everywhere** (OpenTK 4 / GLFW `NativeWindow`). The **Direct3D (SlimDX)** and
  **software (WinForms)** renderers were **dropped** (no .NET 10 packages). `COpenGL` is the only renderer.
- **Audio: PortAudio** (`PortAudioSharp2`) for playback **and** record. **DirectSound / GStreamer / OpenAL
  were dropped.**
- **Images/fonts/text: SkiaSharp** replaces `System.Drawing.Common` (GDI+). `System.Drawing.Common` is
  **removed from the project entirely** (can't be used even by accident); only `System.Drawing.Primitives`
  types (`Color`, `Rectangle`, `PointF`, …) remain. All image loading, font rasterization, the cover
  generator, base64 images and webcam snapshots use SkiaSharp.
- **Audio/video decode: Acinerella + FFmpeg** (ported to FFmpeg 6 on Linux; built from source).
- **Webcam: AForge/DirectShow dropped.** Linux gets a **V4L2** backend (ffmpeg subprocess); other OSes
  currently route to `CNullWebcam` → **no webcam on Windows/macOS in this build yet** (avatar photos).
- **Webserver/remote control: rewritten on ASP.NET Core (Kestrel)** instead of WCF. REST endpoints in
  `CWebservice.MapEndpoints`; still off by default (`Config.Server.ServerActive`).
- **Party modes compiled at runtime with Roslyn** (`Microsoft.CodeAnalysis.CSharp`) instead of CodeDom
  (unsupported on .NET 10). Their sources are bundled next to the mode data at build time.
- **Single codebase, OS-conditional define:** `WIN` on Windows, `LINUX` otherwise, plus additive `MACOS`
  on OSX — chosen per build OS via `MSBuild::IsOSPlatform`. Windows-only code lives behind `#if WIN`.
- **Input:** `VocaluxeLib.Keys` is a custom drop-in enum replacing `System.Windows.Forms.Keys` (same int
  values). Gamepad via GLFW joystick; WiiMote via hidapi with a robust cross-platform loader.

## Windows runtime bring-up (fullscreen freeze fix + startup splash)
Running the port on real Windows surfaced two issues, both now fixed on this branch:
- **Fullscreen present freeze (Windows-only):** exclusive GLFW fullscreen — and any borderless window that
  *exactly* covers the monitor — gets promoted by Windows to fullscreen-exclusive present, which froze our
  manually-driven `SwapBuffers` loop (the visible front buffer stuck on the first frame, even though init
  and the render loop were actually healthy). Known upstream behavior (OpenTK #1631, SDL3 #12791); Linux and
  macOS use a different present path and are unaffected. Fix: on Windows the fullscreen window is sized
  **1px taller than the monitor** so it stays on the composited DWM path (`CDrawBaseWindows._EnterFullScreen`,
  `#if WIN`); Linux/macOS keep native `WindowState.Fullscreen`.
- **Startup splash + loading progress bar:** `OutputType=WinExe` (no stray console window on launch) plus a
  cross-platform OpenGL splash (`CDrawBase.ShowSplash`/`UpdateSplash`, driven from `Program._Run`) shown
  during the ~15s init — logo on the loading-screen background + **progress bar** + status text, with the
  slow party-mode compile reporting per-mode progress so the bar advances instead of looking frozen. This
  restores the splash the port had dropped together with WinForms (`CSplashScreen.cs`); the **progress bar is
  the one deliberate enhancement** over the original (see "Migration philosophy"). Both are cross-platform
  (the splash shows on Linux too).

## Platform testing status
| OS | Builds | Unit tests | Runs / played by hand |
|----|--------|-----------|------------------------|
| **Linux** | ✅ | ✅ | ✅ end-to-end (singing, party modes, webcam, gamepad, webserver, AppImage) |
| **Windows** | ✅ (CI) | ✅ (CI) | ✅ played by hand (singing, party mode) on real hardware after the fullscreen-present fix |
| **macOS** | ✅ (CI) | ✅ (CI) | ❌ **not yet** (experimental) |

CI itself cannot exercise the GUI/audio/native rendering (no display/GPU/sound on runners), so the *CI*
signal is "compiles + units green", not "runs"; the hardware runs above were done by hand. Downloadable test
builds exist precisely so others can validate on real hardware (see below).

## Test coverage
We rely mostly on the **existing upstream test suite** (migrated to .NET 10): 218 tests covering only
**platform-independent `VocaluxeLib` logic** — XML (de)serialization, party-mode-challenge round logic,
combinatorics, logging. These are a good regression net and are what makes the Windows/macOS CI meaningful
(compile + logic green), **but they barely touch what the port changed** — they pass on every OS precisely
because they exercise code we mostly did not modify.

We added **3 headless tests** (`Tests/Vocaluxe/CImagingTest.cs`) for the **GDI+ → SkiaSharp** migration —
the biggest and riskiest change: a `CBase64Image` decode/encode round-trip and a SkiaSharp text-rasterization
check. Besides exercising our wrapper, they prove the native **libSkiaSharp loads and renders on all three OS**
in CI (the font path is where a real glyph bug was fixed during the port). Total is now **221**.

Still **not** unit-tested (need a real display/GPU/sound, so validated by hand on Linux instead): actual GL
rendering output, audio playback (PortAudio), the GLFW window, the webserver at runtime (endpoints marshal to
the game's main thread), native interop loading, input and webcam. Possible follow-ups: webserver endpoint
tests via ASP.NET `WebApplicationFactory` (needs main-thread stubbing) and a profile-migration regression
test (the `ID == Guid.Empty` fix). The unknown-enum-value fallback is already covered by an existing test.

## SonarCloud quality gate — expect it to fail on this PR (by nature, not by neglect)
We previewed SonarCloud's C# rules locally (the `SonarAnalyzer.CSharp` Roslyn analyzers, same rules Sonar
runs). The **new-code quality gate will almost certainly fail**, but for reasons inherent to a one-shot
platform port rather than real defects — please judge accordingly:
- **Coverage (≥80% on new code): unreachable.** The port is mostly GUI/OpenGL/audio/native-interop/webserver
  code that can't be unit-tested without a display, GPU and sound device — so new-code coverage is very low
  (single-digit %). No amount of work changes that; it needs a gate waiver or manual judgement.
- **Security Hotspots Reviewed (100%): not something we can do.** The hotspots (binding Kestrel to
  `0.0.0.0`, native P/Invoke) need to be marked "reviewed/safe" in *your* SonarCloud project UI — we have no
  access to it. They are intended/benign (the server is off by default).
- **Duplicated lines (≤3%): marginally over** (~repeated platform/draw/screen patterns). Not worth a risky
  refactor for a few tenths of a percent.
- **Reliability rating (A): the only code-relevant one.** The few "bug"-flagged items are false positives
  on intentional code (e.g. `S4275` on the deliberately-redefined `CFonts.IsInitialized` getter). **No real
  new bugs were introduced** (verified). Code smells on new code are minimal (and the empty best-effort
  catch blocks are now documented).

In short: the gate metrics here are maintainer-side calls for a PR of this scope, not blockers we can or
should chase down in code.

## CI & downloadable test builds
**Scope note:** this CI is deliberately *minimal* — just enough to verify our cross-platform base (does it
build and do the unit tests pass on all three OS, plus downloadable test builds). We intentionally did **not**
build out a full CI/release pipeline (nightly/release/SonarCloud/deploy), because a dedicated GitHub Actions
migration is already in flight upstream in **[PR #793](https://github.com/Vocaluxe/Vocaluxe/pull/793)**. We
did not want to step on that effort — we only cover what our port needs. If #793 lands, our workflow can be
folded into / replaced by it.

A **single workflow** (`.github/workflows/ci.yml`) with two stages, so there is just one Actions entry:
- **`build-test`** — build + `dotnet test` matrix on `ubuntu/windows/macos-latest` (concurrency cancel,
  NuGet cache, dorny/test-reporter job summary). Catches Windows/macOS build breaks. This is the **gate**.
- **`package-*`** — per-OS **self-contained** artifacts, each `needs: build-test` (so they run **only after
  the tests are green**) and gated to the feature branch / manual dispatch (not every PR):
  - `Vocaluxe-linux-x86_64` — AppImage (ubuntu-22.04 baseline, bundles FFmpeg/PortAudio/fontconfig).
  - `Vocaluxe-windows-x64` — self-contained zip; native C++ DLLs (PitchTracker, acinerella + FFmpeg 4.x,
    portaudio) **reused from the official Vocaluxe Windows nightly** (ABI-compatible, unchanged P/Invoke).
  - `Vocaluxe-macos-arm64` — **experimental**; PitchTracker/acinerella built from source with clang,
    FFmpeg (pinned `ffmpeg@4`) + PortAudio from Homebrew, made relocatable via `dylibbundler`
    (`@loader_path`). SkiaSharp macOS native added via an OSX-only package reference.

## Build & versioning (the SDK migration broke version stamping — now restored)
The .NET Framework project stamped the assembly version at build time via MSBuild **Pre/PostBuildEvents**
(`.build/winPreBuild.ps1` + `linuxPreBuildEvent.sh`, fed by `versionNames.json`) — which AppVeyor triggers
indirectly whenever it builds `Vocaluxe.sln`. **SDK-style projects do not run those build events**, so the
migration silently dropped version stamping: every build reported `Vocaluxe 'Not Versioned' 0.0.0`. That is
the *release-versioning pipeline being broken*, not just a cosmetic default.

Restored it the SDK-native way: a `_StampGitVersion` target in `Vocaluxe.csproj` runs
`git describe --tags --long --always` before compile and generates the `AssemblyTitle` /
`AssemblyInformationalVersion` / `AssemblyVersion` / `AssemblyFileVersion` attributes via `WriteCodeFragment`
(the matching static attributes were removed from `Properties/AssemblyInfo.cs`). It falls back to the old
"Not Versioned" strings when git is unavailable (e.g. a source tarball), so the build never depends on git.
The Package workflow checks out with `fetch-depth: 0` so tags are present and artifacts get a real version.

Follow-ups: the legacy `.build/{win,linux}*Build*.{ps1,sh}` scripts and `versionNames.json` are now
**superseded** (nothing in the SDK build invokes them) and can be removed — kept for now to minimise upstream
churn. The human-readable codename from `versionNames.json` (e.g. "Blue Sunrise") is not reproduced by the
target; current tags have no codename anyway, but re-add a lookup if the maintainer wants it.

## Behavioral / operational changes users should know
- **Logs** now go to the writable data folder (`~/.config/Vocaluxe/Logs` on Linux), **not** next to the
  exe — required so the read-only AppImage mount works. (Previously `<exe>/Logs`.)
- **Config/data** on Linux live under `~/.config/Vocaluxe/` (`Config.xml`, `Profiles/`, `Songs/`,
  `Playlists/`, DBs, `Logs/`) — derived from the OS data folder, identical regardless of launch method, so
  AppImage and a plain build share the same data and survive updates.
- `Program.Main` **sets the working directory to the exe directory** at startup so relative `Output/`
  paths resolve no matter how the app is launched (AppImage AppRun, desktop shortcut, other CWD).
- Native libs load by file name (Linux `*.so`, Windows `*.dll`, macOS mangled `*.dll.dylib`/`*.dylib`).

## Caveats / things to watch
- **Windows regressions vs the current release:** OpenGL instead of Direct3D, PortAudio instead of
  DirectSound (both exercised by hand on Windows — singing + party mode work), and **no webcam** (AForge
  dropped — still unevaluated).
- **Feature stubs introduced by the port** (each fails gracefully / is documented in code):
  - **Song-folder picker dialog is a no-op** — it used WinForms' `FolderBrowserDialog`; it now logs an
    error and returns false. Song folders must be edited in `Config.xml` until a portable folder dialog is
    wired up. (`CScreenOptions._OpenSongFolderDialog`)
  - **Webserver HTTPS is not supported** — the remote control falls back to HTTP. (`CVocaluxeServer`)
  - **No crash reporter** — the WinForms `CReporter` was removed; unhandled errors are logged but there is
    no cross-platform reporter UI yet. (`Program.cs`)
  - Pre-existing `NotImplementedException`s (cover-DB schema upgrade, song-writer note types) are upstream,
    not introduced here.
- **macOS is arm64-only and unsigned** → Gatekeeper blocks double-click (right-click → Open, or
  `xattr -dr com.apple.quarantine`). Intel (osx-x64) not built yet. `ARCH_X64` is hardcoded in the
  defines (semantically wrong on Apple Silicon but compiles; only matters for the Windows `libs/…/x64`
  path logic, unused on macOS).
- Theme background **videos are not in the repo** (downloaded separately from VocaluxeDependencies), so
  the CI test builds ship without them — menus work, just without background video.
- `WebserverInitalConfig/` (IIS helper) is **Windows-only** and excluded from the Linux build.

## Sync with `develop`
- `develop` advanced by 3 commits since our base. Two are **included**:
  - **#879 (Remove ReloadSongMethod)** — cherry-picked (clean); the method is gone from our code.
  - **#878 (multiple/multi-value song tags, UltraStar US format V1+V2)** — cherry-picked and conflict-resolved
    on top of the port (8 files): took #878's feature (file-path field renames `MP3FileName`→`Audio`,
    `Cover`, `Video`, …, multi-value `Creators`/`Languages`/`Genres`/`Tags`) while keeping the port's
    SkiaSharp cover path, Kestrel webserver, cross-platform `Path.Combine`, and structured logging. Build +
    all 221 tests green. (#878's `Id`/`Creator` references that came from #877 were back-adapted to our old
    names.)
- **#877 (R# clean-up) is intentionally NOT taken**, even though it has now landed on `develop`. It is purely cosmetic on the *old* codebase
  (≈5.7k added braces, ≈2.5k `var`, formatting, usings, dep/LangVersion bumps we already have) with **no
  behavioral changes**. The one non-cosmetic part is naming-convention renames (`ID→Id`, `BPM→Bpm`,
  `Finish→End`) — functionally identical; our code still uses the old names. **Optional follow-up:** align
  to those names if the maintainer wants style consistency (not required for correctness).

## Suggested reviewer how-to
```bash
# Linux build + run
export DOTNET_ROOT="$HOME/.dotnet"; export PATH="$HOME/.dotnet:$PATH"
./.build/build-linux.sh && ./dist/Vocaluxe.sh
# or grab a self-contained AppImage / Windows zip / macOS build from the CI "Package" run artifacts
dotnet test Tests/Tests.csproj -c Release      # 218 tests, all OS
```

## Open / deferred (not in this PR)
- Real macOS hands-on validation. (Windows is now validated by hand — singing + party mode; remaining
  Windows gap is webcam.)
- macOS Intel (osx-x64) build; macOS code signing/notarization.
- Webcam on Windows/macOS; mic-sensitivity option; song-folder union.
- Optional: apply develop's `ID→Id`/`BPM→Bpm`/`Finish→End` naming for style parity.


byte55:
#877  landed on develop after this port branch was already underway. I left it out of this PR: beyond the cosmetic reformatting it also does naming-convention renames (ID→Id, BPM→Bpm, Finish→End), and I kept the original name here. The port already does the non-cosmetic modernization independently (dependency updates, latest C#, using clean-up) as part of the .NET 10 migration. I did separately cherry-pick the functional PRs that landed in the meantime, #878 and #879. 
That's the only reason GitHub marks the branch as not mergeable: it still uses the property names that #877 renamed. Happy to rebase onto develop and adopt those renames if you'd like it to merge cleanly — just say the word.

poupounetjoyeux:
Little Bigbang PR.. 👌
BTW #880 and one more for audio side (already developped but based on it so waiting to push it) remove the need of Acinerella if it could be usefull

byte55:
> Little Bigbang PR.. 👌 BTW #880 and one more for audio side (already developped but based on it so waiting to push it) remove the need of Acinerella if it could be usefull

Please don't remove it. I had some serious problems while porting this one :sweat_smile: 

I can try to merge the remaining PR as soon as they are merged into develop :ok_hand:

poupounetjoyeux:
We want to remove it to be able to deal with latest FFmpeg versions (currently my code runs with 8.1.2)

FFmpeg.Autogen is fully compatible with .NET 10 and allows us more flexibility with FFmpeg:
- HW acceleration for video that is in my pipe and that I should finish today
- The add of auto 16bits resampling to fix a compatibility issue with port audio buffers (don't know if PortAudioSharp 2 is also impacted but if it uses unmanaged DLL binding it is probably 😅)

So yes if they are merged before your please backport them 👍

byte55:
Until upstream has capacity for this again, I'll keep working in my fork with only my own Linux machine as the target, so further commits here won't carry Windows/macOS validation anymore. Should I close this PR in the meantime so you don't get pinged for every push? Happy to come back with a fresh one rebased on top of the FFmpeg work later or to leave it open, whichever you prefer.

flokuep:
Sorry. I'm really busy, I don't have enough time to do a proper review of the code. Might it be possible to split this up into smaller prs? Otherwise you might better continue with your fork for now 🙃

